PR #11, opened 2021-02-06
Merge new component - Initial commit 

DurgNomis-drol:
I dont know if i have made the pull request correct - Please tell me if so and i will try to fix it

This will break older configurations.

**Thing's working:**
- Token, Timer and Next alarm sensor for each device.

**Thing's not working**
- Can only retrieve one device as of now. Homegraph doesn't have a way to retrieve an IP of each device, so we need to find another way to do that. See API.py to change the IP for testing.
- API retrieves data every 15 seconds. But i can't figure out how to update entities. Entities only updates when you reload the integration.

**Thing's to talk about**
- Should each alarm set be a separate entity or as attributes in one entity?
- Should each timer set be a separate entity or as attributes in one entity?

I will try my best to help with maintaining this integration, but my skills as a programmer is limited.

### DIFF
--- a/custom_components/ha-glocaltokens/__init__.py
+++ b/custom_components/ha-glocaltokens/__init__.py
@@ -23,10 +23,8 @@ from .const import CONF_PASSWORD
 from .const import CONF_USERNAME
 from .const import DOMAIN
 from .const import PLATFORMS
-from .const import SCAN_INTERVAL
+from .const import TIME_BETWEEN_UPDATES
 from .const import STARTUP_MESSAGE
-
-SCAN_INTERVAL = timedelta(seconds=SCAN_INTERVAL)
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
@@ -80,16 +78,16 @@ class GlocaltokensDataUpdateCoordinator(DataUpdateCoordinator):
         self.api = client
         self.platforms = []
 
-        super().__init__(hass, _LOGGER, name=DOMAIN, update_interval=SCAN_INTERVAL)
+        super().__init__(hass, _LOGGER, name=DOMAIN, update_interval=TIME_BETWEEN_UPDATES)
 
     def _update_data(self) -> dict:
         """Fetch local auth tokens from Google via sync functions."""
-        return self.api.get_google_devices_json()
+        return self.api.get_google_devices_information()
 
     async def _async_update_data(self):
         """Update data via library."""
         try:
-            async with timeout(15):
+            async with timeout(10):
                 return await self.hass.async_add_executor_job(self._update_data)
         except Exception as exception:
             raise UpdateFailed(f"Invalid response from API: {exception}") from exception
--- a/custom_components/ha-glocaltokens/__init__.py
+++ b/custom_components/ha-glocaltokens/__init__.py
@@ -23,8 +23,10 @@ from .const import CONF_PASSWORD
 from .const import CONF_USERNAME
 from .const import DOMAIN
 from .const import PLATFORMS
-from .const import TIME_BETWEEN_UPDATES
+from .const import SCAN_INTERVAL
 from .const import STARTUP_MESSAGE
+
+SCAN_INTERVAL = timedelta(seconds=SCAN_INTERVAL)
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
--- a/custom_components/ha-glocaltokens/__init__.py
+++ b/custom_components/ha-glocaltokens/__init__.py
@@ -80,7 +80,7 @@ class GlocaltokensDataUpdateCoordinator(DataUpdateCoordinator):
         self.api = client
         self.platforms = []
 
-        super().__init__(hass, _LOGGER, name=DOMAIN, update_interval=TIME_BETWEEN_UPDATES)
+        super().__init__(hass, _LOGGER, name=DOMAIN, update_interval=SCAN_INTERVAL)
 
     def _update_data(self) -> dict:
         """Fetch local auth tokens from Google via sync functions."""
--- a/custom_components/ha-glocaltokens/api.py
+++ b/custom_components/ha-glocaltokens/api.py
@@ -4,10 +4,12 @@ import requests
 import json
 import aiohttp
 
+from .utils import convert_from_ms_to_s
+
 from datetime import datetime
 
 from .const import API_RETURNED_UNKNOWN
-from .const import PORT, ENDPOINT, HEADERS, HEADER_CAST_LOCAL_AUTH
+from .const import PORT, API_ENDPOINT_ALARMS, HEADERS, HEADER_CAST_LOCAL_AUTH
 from .const import FIRE_TIME, DATE_TIME, DURATION, ORIGINAL_DURATION, LOCAL_TIME
 from .const import SHOW_TIME_ONLY, SHOW_DATE_AND_TIME, SHOW_DATE_TIMEZONE
 from .const import GLOCALTOKENS_ALARMS
@@ -19,10 +21,7 @@ from homeassistant.const import HTTP_OK
 
 TIMEOUT = 10
 
-
 _LOGGER: logging.Logger = logging.getLogger(__package__)
-
-HEADERS = {"Content-type": "application/json; charset=UTF-8"}
 
 
 class GlocaltokensApiClient:
@@ -58,34 +57,31 @@ class GlocaltokensApiClient:
         """Generate random android_id"""
         return self._client._get_android_id()
 
-    def format_timer_information(timer):
-        timestamp_ms = timer[FIRE_TIME]
-        timestamp = timestamp_ms / 1000
-        humantime = datetime.fromtimestamp(timestamp).strftime(SHOW_TIME_ONLY)
-        timer[DATE_TIME] = humantime
+    def format_timer_information(self, timer):
+        timestamp = convert_from_ms_to_s(timer[FIRE_TIME])
+        timer[DATE_TIME] = datetime.fromtimestamp(timestamp).strftime(SHOW_TIME_ONLY)
 
-        duration_ms = timer[ORIGINAL_DURATION]
-        duration = duration_ms / 1000
-        humanduration = datetime.utcfromtimestamp(duration).strftime(SHOW_TIME_ONLY)
-        timer[DURATION] = humanduration
+        duration = convert_from_ms_to_s(timer[ORIGINAL_DURATION])
+        timer[DURATION] = datetime.utcfromtimestamp(duration).strftime(SHOW_TIME_ONLY)
         return timer
 
-    def format_alarm_information(alarm):
-        timestamp_ms = alarm[FIRE_TIME]
-        timestamp = timestamp_ms / 1000
-        humantime = datetime.utcfromtimestamp(timestamp).strftime(SHOW_DATE_TIMEZONE)
-        localtime = datetime.fromtimestamp(timestamp).strftime(SHOW_DATE_AND_TIME)
-        alarm[DATE_TIME] = humantime
-        alarm[LOCAL_TIME] = localtime
+    def format_alarm_information(self, alarm):
+        timestamp = convert_from_ms_to_s(alarm[FIRE_TIME])
+        alarm[DATE_TIME] = datetime.utcfromtimestamp(timestamp).strftime(SHOW_DATE_TIMEZONE)
+        alarm[LOCAL_TIME] = datetime.fromtimestamp(timestamp).strftime(SHOW_DATE_AND_TIME)
         return alarm
 
-    def create_url(ip):
-        url = 'https://'+ ip +':'+ str(PORT) + ENDPOINT
+    def create_url(self, ip, api_endpoint):
+        url = 'https://{ip}:{port}/{endpoint}'.format(
+            ip=ip,
+            port=str(PORT),
+            endpoint=api_endpoint
+        )
         return url
 
-    def get_alarms_and_timers_from(ip, token):
+    def get_alarms_and_timers_from(self, ip, token, endpoint):
 
-        url = GlocaltokensApiClient.create_url(ip)
+        url = self.create_url(ip, endpoint)
 
         HEADERS[HEADER_CAST_LOCAL_AUTH] = token
 
@@ -97,15 +93,16 @@ class GlocaltokensApiClient:
 
         devices = self._client.get_google_devices_json()
 
-        for x in range(len(devices)):
-            #local_token = next(item[GLOCALTOKENS_TOKEN] for item in devices)
-            local_token = devices[1][GLOCALTOKENS_TOKEN]
+        for device in range(len(devices)):
+            device_array = devices[1]
 
-            response = GlocaltokensApiClient.get_alarms_and_timers_from('192.168.0.205', local_token) # IP
+            # local_token = device[GLOCALTOKENS_TOKEN]
+            local_token = device_array[GLOCALTOKENS_TOKEN] # ONLY BECAUSE WE DONT HAVE DISCOVERY YET - I manuly select device so i can test
+
+            response = self.get_alarms_and_timers_from('192.168.0.205', local_token, API_ENDPOINT_ALARMS) # IP
 
             if response.status_code != HTTP_OK:
                 _LOGGER.error("API returned {}".format(response.status_code))
-                #_LOGGER.error(local_token)
                 return devices
 
             result = response.json()
@@ -114,12 +111,11 @@ class GlocaltokensApiClient:
                 _LOGGER.error(API_RETURNED_UNKNOWN)
                 return devices
 
-            devices[x][GLOCALTOKENS_TIMERS] = result[GLOCALTOKENS_TIMERS]
-            for y in range(len(result[GLOCALTOKENS_TIMERS])):
-                devices[x][GLOCALTOKENS_TIMERS][y] = GlocaltokensApiClient.format_timer_information(result[GLOCALTOKENS_TIMERS][y])
+            timers = result[GLOCALTOKENS_TIMERS]
+            devices[device][GLOCALTOKENS_TIMERS] = [self.format_timer_information(timer) for timer in timers]
 
-            devices[x][GLOCALTOKENS_ALARMS] = result[GLOCALTOKENS_ALARMS]
-            for z in range(len(result[GLOCALTOKENS_ALARMS])):
-                devices[x][GLOCALTOKENS_ALARMS][z] = GlocaltokensApiClient.format_timer_information(result[GLOCALTOKENS_ALARMS][z])
+            alarms = result[GLOCALTOKENS_ALARMS]
+            devices[device][GLOCALTOKENS_ALARMS] = [self.format_alarm_information(alarm) for alarm in alarms]
+
 
         return devices
--- a/custom_components/ha-glocaltokens/api.py
+++ b/custom_components/ha-glocaltokens/api.py
@@ -58,33 +58,50 @@ class GlocaltokensApiClient:
         """Generate random android_id"""
         return self._client._get_android_id()
 
-    def format_timer_information(device_timers, timer):
-        timestamp_ms = result[GLOCALTOKENS_TIMERS][timer][FIRE_TIME]
+    def format_timer_information(timer):
+        timestamp_ms = timer[FIRE_TIME]
         timestamp = timestamp_ms / 1000
         humantime = datetime.fromtimestamp(timestamp).strftime(SHOW_TIME_ONLY)
-        devices[x][GLOCALTOKENS_TIMERS][y][DATE_TIME] = humantime
+        timer[DATE_TIME] = humantime
 
-        duration_ms = result[GLOCALTOKENS_TIMERS][y][ORIGINAL_DURATION]
+        duration_ms = timer[ORIGINAL_DURATION]
         duration = duration_ms / 1000
         humanduration = datetime.utcfromtimestamp(duration).strftime(SHOW_TIME_ONLY)
-        devices[x][GLOCALTOKENS_TIMERS][y][DURATION] = humanduration
+        timer[DURATION] = humanduration
+        return timer
 
+    def format_alarm_information(alarm):
+        timestamp_ms = alarm[FIRE_TIME]
+        timestamp = timestamp_ms / 1000
+        humantime = datetime.utcfromtimestamp(timestamp).strftime(SHOW_DATE_TIMEZONE)
+        localtime = datetime.fromtimestamp(timestamp).strftime(SHOW_DATE_AND_TIME)
+        alarm[DATE_TIME] = humantime
+        alarm[LOCAL_TIME] = localtime
+        return alarm
+
+    def create_url(ip):
+        url = 'https://'+ ip +':'+ str(PORT) + ENDPOINT
+        return url
+
+    def get_alarms_and_timers_from(ip, token):
+
+        url = GlocaltokensApiClient.create_url(ip)
+
+        HEADERS[HEADER_CAST_LOCAL_AUTH] = token
+
+        response = requests.get(url, headers=HEADERS, verify=False, timeout=TIMEOUT)
+
+        return response
 
     def get_google_devices_information(self):
 
         devices = self._client.get_google_devices_json()
 
-        _LOGGER.error("Fetching new data...")
-
         for x in range(len(devices)):
             #local_token = next(item[GLOCALTOKENS_TOKEN] for item in devices)
             local_token = devices[1][GLOCALTOKENS_TOKEN]
 
-            url = 'https://192.168.0.205:'+ str(PORT) + ENDPOINT
-
-            HEADERS[HEADER_CAST_LOCAL_AUTH] = local_token
-
-            response = requests.get(url, headers=HEADERS, verify=False, timeout=TIMEOUT)
+            response = GlocaltokensApiClient.get_alarms_and_timers_from('192.168.0.205', local_token) # IP
 
             if response.status_code != HTTP_OK:
                 _LOGGER.error("API returned {}".format(response.status_code))
@@ -97,21 +114,12 @@ class GlocaltokensApiClient:
                 _LOGGER.error(API_RETURNED_UNKNOWN)
                 return devices
 
-            _LOGGER.error(result)
-
             devices[x][GLOCALTOKENS_TIMERS] = result[GLOCALTOKENS_TIMERS]
             for y in range(len(result[GLOCALTOKENS_TIMERS])):
-                format_timer_information(result[GLOCALTOKENS_TIMERS], y)
-
+                devices[x][GLOCALTOKENS_TIMERS][y] = GlocaltokensApiClient.format_timer_information(result[GLOCALTOKENS_TIMERS][y])
 
             devices[x][GLOCALTOKENS_ALARMS] = result[GLOCALTOKENS_ALARMS]
             for z in range(len(result[GLOCALTOKENS_ALARMS])):
-
-                timestamp_ms = result[GLOCALTOKENS_ALARMS][z][FIRE_TIME]
-                timestamp = timestamp_ms / 1000
-                humantime = datetime.utcfromtimestamp(timestamp).strftime(SHOW_DATE_TIMEZONE)
-                localtime = datetime.fromtimestamp(timestamp).strftime(SHOW_DATE_AND_TIME)
-                devices[x][GLOCALTOKENS_ALARMS][z][DATE_TIME] = humantime
-                devices[x][GLOCALTOKENS_ALARMS][z][LOCAL_TIME] = localtime
+                devices[x][GLOCALTOKENS_ALARMS][z] = GlocaltokensApiClient.format_timer_information(result[GLOCALTOKENS_ALARMS][z])
 
         return devices
--- a/custom_components/ha-glocaltokens/api.py
+++ b/custom_components/ha-glocaltokens/api.py
@@ -1,10 +1,16 @@
 """Sample API Client."""
 import logging
-
+import requests
+import json
 import aiohttp
-from glocaltokens.client import GLocalAuthenticationTokens
 
+from datetime import datetime
+
+from .const import GLOCALTOKENS_ALARMS
+from .const import GLOCALTOKENS_TOKEN
+from .const import GLOCALTOKENS_TIMERS
 from .exceptions import InvalidMasterToken
+from glocaltokens.client import GLocalAuthenticationTokens
 
 TIMEOUT = 10
 
@@ -46,3 +52,57 @@ class GlocaltokensApiClient:
     def get_android_id(self):
         """Generate random android_id"""
         return self._client._get_android_id()
+
+    def get_google_devices_information(self):
+
+        devices = self._client.get_google_devices_json()
+
+        _LOGGER.error("Fetching new data...")
+
+        for x in range(len(devices)):
+            #local_token = next(item[GLOCALTOKENS_TOKEN] for item in devices)
+            local_token = devices[1][GLOCALTOKENS_TOKEN]
+
+            url = 'https://192.168.0.205:8443/setup/assistant/alarms'
+            header = {'cast-local-authorization-token': local_token,
+                      'content-type': 'application/json'}
+
+            response = requests.get(url, headers=header, verify=False, timeout=TIMEOUT)
+
+            if response.status_code != 200:
+                _LOGGER.error("API returned {}".format(response.status_code))
+                #_LOGGER.error(local_token)
+                return devices
+
+            result = response.json()
+
+            if GLOCALTOKENS_TIMERS not in result and GLOCALTOKENS_ALARMS not in result:
+                _LOGGER.error("API returned unknown json structure")
+                return devices
+
+            _LOGGER.error(result)
+
+            devices[x][GLOCALTOKENS_TIMERS] = result[GLOCALTOKENS_TIMERS]
+            for y in range(len(result[GLOCALTOKENS_TIMERS])):
+
+                timestamp_ms = result[GLOCALTOKENS_TIMERS][y]['fire_time']
+                timestamp = timestamp_ms / 1000
+                humantime = datetime.fromtimestamp(timestamp).strftime('%H:%M:%S')
+                devices[x][GLOCALTOKENS_TIMERS][y]['date_time'] = humantime
+
+                duration_ms = result[GLOCALTOKENS_TIMERS][y]['original_duration']
+                duration = duration_ms / 1000
+                humanduration = datetime.utcfromtimestamp(duration).strftime('%H:%M:%S')
+                devices[x][GLOCALTOKENS_TIMERS][y]['duration'] = humanduration
+
+            devices[x][GLOCALTOKENS_ALARMS] = result[GLOCALTOKENS_ALARMS]
+            for z in range(len(result[GLOCALTOKENS_ALARMS])):
+
+                timestamp_ms = result[GLOCALTOKENS_ALARMS][z]['fire_time']
+                timestamp = timestamp_ms / 1000
+                humantime = datetime.utcfromtimestamp(timestamp).strftime('%Y-%m-%dT%H:%M:%S.%fZ%Z')
+                localtime = datetime.fromtimestamp(timestamp).strftime('%Y-%m-%d %H:%M:%S')
+                devices[x][GLOCALTOKENS_ALARMS][z]['date_time'] = humantime
+                devices[x][GLOCALTOKENS_ALARMS][z]['local_time'] = localtime
+
+        return devices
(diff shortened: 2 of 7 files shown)


custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +52,53 @@ def get_google_devices_json(self):
...
+            for z in range(len(result[GLOCALTOKENS_ALARMS])):
+
+                timestamp_ms = result[GLOCALTOKENS_ALARMS][z]['fire_time']
+                timestamp = timestamp_ms / 1000
+                humantime = datetime.utcfromtimestamp(timestamp).strftime('%Y-%m-%dT%H:%M:%S.%fZ%Z')
+                localtime = datetime.fromtimestamp(timestamp).strftime('%Y-%m-%d %H:%M:%S')
+                devices[x][GLOCALTOKENS_ALARMS][z]['date_time'] = humantime
+                devices[x][GLOCALTOKENS_ALARMS][z]['local_time'] = localtime
+
+        return devices
```
leikoilja, line 104: That's a really long method, which is a bit hard to read. Do you mind us splitting that into smaller subfunctions for easier readability? 👍 

I will take a look at extracting device IPs using `glocaltokens` client (See https://github.com/leikoilja/ha-glocaltokens/issues/16#issuecomment-776796188), so this get's a bit cleaner.

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +52,53 @@ def get_google_devices_json(self):
...
+            #local_token = next(item[GLOCALTOKENS_TOKEN] for item in devices)
+            local_token = devices[0][GLOCALTOKENS_TOKEN] # Else x
+
+            url = 'https://IP OF DEVICE:8443/setup/assistant/alarms' # ONLY FOR TESTING PURPOSE - Until we find a way to retrieve ip of each device.
+            header = {'cast-local-authorization-token': local_token,
+                      'content-type': 'application/json'}
+
+            response = requests.get(url, headers=header, verify=False, timeout=TIMEOUT)
+
+            if response.status_code != 200:
```
leikoilja, line 70: The constants like request status `200` is a constant and we can access those from pre-deficed HA constants like `from homeassistant.const import HTTP_OK`

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +52,53 @@ def get_google_devices_json(self):
...
+    def get_google_devices_information(self):
+
+        devices = self._client.get_google_devices_json()
+
+        for x in range(len(devices)):
+            #local_token = next(item[GLOCALTOKENS_TOKEN] for item in devices)
+            local_token = devices[0][GLOCALTOKENS_TOKEN] # Else x
+
+            url = 'https://IP OF DEVICE:8443/setup/assistant/alarms' # ONLY FOR TESTING PURPOSE - Until we find a way to retrieve ip of each device.
+            header = {'cast-local-authorization-token': local_token,
```
leikoilja, line 65: Let's have those headers as constants in the `const.py` file

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +52,53 @@ def get_google_devices_json(self):
...
+            result = response.json()
+
+            if GLOCALTOKENS_TIMERS not in result and GLOCALTOKENS_ALARMS not in result:
+                _LOGGER.error("API returned unknown json structure")
+                return devices
+
+            devices[x][GLOCALTOKENS_TIMERS] = result[GLOCALTOKENS_TIMERS]
+            for y in range(len(result[GLOCALTOKENS_TIMERS])):
+
+                timestamp_ms = result[GLOCALTOKENS_TIMERS][y]['fire_time']
```
leikoilja, line 84: Maybe we should move those parameters as `'fire_time'` and `'original_duration'` etc as constants too

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +52,53 @@ def get_google_devices_json(self):
...
+                duration = duration_ms / 1000
+                humanduration = datetime.utcfromtimestamp(duration).strftime('%H:%M:%S')
+                devices[x][GLOCALTOKENS_TIMERS][y]['duration'] = humanduration
+
+            devices[x][GLOCALTOKENS_ALARMS] = result[GLOCALTOKENS_ALARMS]
+            for z in range(len(result[GLOCALTOKENS_ALARMS])):
+
+                timestamp_ms = result[GLOCALTOKENS_ALARMS][z]['fire_time']
+                timestamp = timestamp_ms / 1000
+                humantime = datetime.utcfromtimestamp(timestamp).strftime('%Y-%m-%dT%H:%M:%S.%fZ%Z')
```
leikoilja, line 99: The time format should also be constant so it would be easier to change if needed. I also wonder if HA don't have any pre-defined one, that would be even better, so we keep it consistent with them

custom_components/ha-glocaltokens/const.py
```diff
@@ -27,13 +30,16 @@
 CONF_MASTER_TOKEN = "master_token"
 
 # Defaults
-DEFAULT_NAME = "GlocalToken"
+DEFAULT_NAME = "GlocalHome" # SUGGESTION - Maybe another name to make it clear that it also includes other things then a token?
```
leikoilja, line 33: That would be great, I all up for it! Do you have any suggestions?

custom_components/ha-glocaltokens/entity.py
```diff
@@ -12,24 +13,11 @@ def __init__(self, coordinator, config_entry):
...
-    @property
-    def unique_id(self):
-        """Return a unique ID to use for this entity."""
-        return self._name
-
     @property
     def device_info(self):
         return {
-            "identifiers": {(DOMAIN, self.unique_id)},
-            "name": NAME,
```
leikoilja, line 24: I am afraid we might need to keep `unique_id` property method untouched. It comes from the integration blueprint and might be used internally in HA, so even though it looks redundant, i would suggest we keep it

custom_components/ha-glocaltokens/entity.py
```diff
@@ -12,24 +13,11 @@ def __init__(self, coordinator, config_entry):
...
             "manufacturer": MANUFACTURER,
         }
-
-    @property
-    def device_state_attributes(self):
-        """Return the state attributes."""
-        return {
-            "id": str(self.unique_id),
-            "integration": DOMAIN,
-        }
```
leikoilja, line 35: Same goes for this method, it might be required for HA. At least it would make us more consistent with the HA base code. Perhaps as the integration grows we can even make use of it

custom_components/ha-glocaltokens/manifest.json
```diff
@@ -1,10 +1,10 @@
...
-  "domain": "ha-glocaltokens",
-  "name": "HA-Glocaltokens",
+  "domain": "glocaltokens",
+  "name": "Google local authentication token fetching (Glocaltokens)",
   "documentation": "https://github.com/leikoilja/ha-glocaltokens",
   "issue_tracker": "https://github.com/leikoilja/ha-glocaltokens/issues",
   "dependencies": [],
   "config_flow": true,
   "codeowners": ["@leikoilja"],
-  "requirements": ["glocaltokens==0.1.4"]
```
leikoilja, line 9: The manifest must have the package version for us to be able to merge it in HACS according to their documentation.

custom_components/ha-glocaltokens/manifest.json
```diff
@@ -1,10 +1,10 @@
 {
-  "domain": "ha-glocaltokens",
-  "name": "HA-Glocaltokens",
```
leikoilja, line 3: I was thinking of having domain and name a bit different from `glocaltokens`, since `glocaltokens` is standalone python package, where this one is HomeAssistant specific integration

leikoilja: Especially now that we are expanding the integration to do more functionality

custom_components/ha-glocaltokens/sensor.py
```diff
@@ -43,3 +62,92 @@ def icon(self):
...
         return "glocaltokens__custom_device_class"
+
+class GlocaltokensAlarmSensor(GlocaltokensEntity):
+    """Representation of a Sensor."""
+
+    def __init__(self, coordinator, entry, device):
+        """Initialize the sensor."""
+        super().__init__(coordinator, entry)
+        self._name = device[GLOCALTOKENS_DEVICE_NAME]
+        self._token = device[GLOCALTOKENS_TOKEN]
```
leikoilja, line 73: If we keep the token in the `GlocaltokensAlarmSensor` do we need the separate `GlocaltokensSensor` sensor holding the same token? We should probably reuse it

custom_components/ha-glocaltokens/sensor.py
```diff
@@ -43,3 +62,92 @@ def icon(self):
...
+        self._alarms = device[GLOCALTOKENS_ALARMS]
+
+        if len(self._alarms) > 0:
+            self._state = self._alarms[0]['fire_time'] / 1000
```
leikoilja, line 77: Same, it would be nice to have all those key words like `fire_time` as a constant, while having this conversion ` / 1000` as a separate method. We can even add a whole separate file like `utils.py` with reusable methods that we can access across the integration

leikoilja (changes requested):
Awesome! That's some really cool work you have done, thanks! 
I have left some minor comments we can discuss, but in general - it's really cool! 💥

custom_components/ha-glocaltokens/__init__.py
```diff
@@ -23,11 +23,9 @@
 from .const import CONF_USERNAME
 from .const import DOMAIN
 from .const import PLATFORMS
-from .const import SCAN_INTERVAL
```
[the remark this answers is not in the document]
DurgNomis-drol: I will change it back, i only changed because of some testing i did.

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: Done, we should always use pre-defind const from homeassistant when possible. I dont know why i did here

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: Done

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: Done and also a lot of  other constants

custom_components/ha-glocaltokens/entity.py
DurgNomis-drol: See sensor.py instead. The reason i have moved it is so each sensor created in HA is listed under each device instead of multiple devices for the same device in the device registry. I dont know if there is a better way to implement it

custom_components/ha-glocaltokens/manifest.json
DurgNomis-drol: I dont know why i have removed it

custom_components/ha-glocaltokens/manifest.json
DurgNomis-drol: my suggestion would be glocalhome

custom_components/ha-glocaltokens/sensor.py
DurgNomis-drol: No, i just forgot to removed from timer and alarm sensor

custom_components/ha-glocaltokens/sensor.py
DurgNomis-drol: i think that is a good idea to separate it into utils.py

custom_components/ha-glocaltokens/entity.py
leikoilja: That's actually a better design! Let's keep it this way

custom_components/ha-glocaltokens/manifest.json
leikoilja: I like it! @DurgNomis-drol, drop it here: https://github.com/leikoilja/ha-glocaltokens/issues/19

leikoilja: Let's not rename it in this PR, but have a whole new PR to rename everything at one once we settle on the name

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: I will start work on this

DurgNomis-drol:
Here is something to work of, i have done everything except dividing the function into smaller ones.

leikoilja:
@DurgNomis-drol feel free to mark conversation that are discussed and set as "Resolve conversation". Also please re-request my review when ready ;)

custom_components/ha-glocaltokens/manifest.json
DurgNomis-drol: Good idea

custom_components/ha-glocaltokens/sensor.py
DurgNomis-drol: We probably should do that to make it easier in the future to add new things

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +57,69 @@ def get_google_devices_json(self):
...
+        timestamp_ms = alarm[FIRE_TIME]
+        timestamp = timestamp_ms / 1000
+        humantime = datetime.utcfromtimestamp(timestamp).strftime(SHOW_DATE_TIMEZONE)
+        localtime = datetime.fromtimestamp(timestamp).strftime(SHOW_DATE_AND_TIME)
+        alarm[DATE_TIME] = humantime
+        alarm[LOCAL_TIME] = localtime
+        return alarm
+
+    def create_url(ip):
+        url = 'https://'+ ip +':'+ str(PORT) + ENDPOINT
```
leikoilja, line 83: Avoid using string casting like this. A more pythonic way would be like this:
```python
url = 'https://{ip}:{port}/{endpoint}'.format(
         ip=ip, 
         port=PORT,
         endpoint=ENDPOINT
)
```
You could also use f-string, but those aren't backwards compatible with python2, so `format` is a safe way to go :)

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +57,69 @@ def get_google_devices_json(self):
...
+
+        response = requests.get(url, headers=HEADERS, verify=False, timeout=TIMEOUT)
+
+        return response
+
+    def get_google_devices_information(self):
+
+        devices = self._client.get_google_devices_json()
+
+        for x in range(len(devices)):
```
leikoilja, line 100: Let's better use `for device in devices:`

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +57,69 @@ def get_google_devices_json(self):
...
+            #local_token = next(item[GLOCALTOKENS_TOKEN] for item in devices)
+            local_token = devices[1][GLOCALTOKENS_TOKEN]
```
leikoilja, line 102: Then here you can just say `local_token = device[GLOCALTOKENS_TOKEN]`

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +57,69 @@ def get_google_devices_json(self):
...
+                #_LOGGER.error(local_token)
+                return devices
+
+            result = response.json()
+
+            if GLOCALTOKENS_TIMERS not in result and GLOCALTOKENS_ALARMS not in result:
+                _LOGGER.error(API_RETURNED_UNKNOWN)
+                return devices
+
+            devices[x][GLOCALTOKENS_TIMERS] = result[GLOCALTOKENS_TIMERS]
```
leikoilja, line 117: Then here you can just say `device[GLOCALTOKENS_TIMERS]` or even better:
```python
timers = result[GLOCALTOKENS_TIMERS]
alarms = result[GLOCALTOKENS_ALARMS]
```

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +57,69 @@ def get_google_devices_json(self):
     def get_android_id(self):
         """Generate random android_id"""
         return self._client._get_android_id()
+
+    def format_timer_information(timer):
```
leikoilja, line 61: You need `self` here as a first argument, because it's a class method.

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +57,69 @@ def get_google_devices_json(self):
...
+        humantime = datetime.fromtimestamp(timestamp).strftime(SHOW_TIME_ONLY)
+        timer[DATE_TIME] = humantime
+
+        duration_ms = timer[ORIGINAL_DURATION]
+        duration = duration_ms / 1000
+        humanduration = datetime.utcfromtimestamp(duration).strftime(SHOW_TIME_ONLY)
+        timer[DURATION] = humanduration
+        return timer
+
+    def format_alarm_information(alarm):
```
leikoilja, line 73: You need `self` here as a first argument, because it's a class method.

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +57,69 @@ def get_google_devices_json(self):
...
+
+            result = response.json()
+
+            if GLOCALTOKENS_TIMERS not in result and GLOCALTOKENS_ALARMS not in result:
+                _LOGGER.error(API_RETURNED_UNKNOWN)
+                return devices
+
+            devices[x][GLOCALTOKENS_TIMERS] = result[GLOCALTOKENS_TIMERS]
+            for y in range(len(result[GLOCALTOKENS_TIMERS])):
+                devices[x][GLOCALTOKENS_TIMERS][y] = GlocaltokensApiClient.format_timer_information(result[GLOCALTOKENS_TIMERS][y])
```
leikoilja, line 119: What are we doing here? Is it just to iterate through all the timers and replace them with human readable format? 
If so, you could achieve the same in a bit cleaner way like so 
`timers = [self.format_timer_information(timer) for timer in timers]`

custom_components/ha-glocaltokens/const.py
```diff
@@ -27,13 +31,42 @@
 CONF_MASTER_TOKEN = "master_token"
 
 # Defaults
-DEFAULT_NAME = "GlocalToken"
+DEFAULT_NAME = "Glocaltokens"
+
+# API
+PORT = 8443
+ENDPOINT = '/setup/assistant/alarms'
```
leikoilja, line 38: Let's name it `API_ENDPOINT_ALARMS`, because as time goes we might introduce more and more API URLs and need to have a consistent way of naming them

custom_components/ha-glocaltokens/manifest.json
```diff
@@ -1,10 +1,10 @@
 {
   "domain": "ha-glocaltokens",
-  "name": "HA-Glocaltokens",
+  "name": "HA-Glocaltokens)",
```
leikoilja, line 3: Please remove `)`

custom_components/ha-glocaltokens/manifest.json
```diff
@@ -1,10 +1,10 @@
...
   "documentation": "https://github.com/leikoilja/ha-glocaltokens",
   "issue_tracker": "https://github.com/leikoilja/ha-glocaltokens/issues",
   "dependencies": [],
   "config_flow": true,
   "codeowners": ["@leikoilja"],
-  "requirements": ["glocaltokens==0.1.4"]
+  "requirements": ["glocaltokens==0.1.4","requests==2.23.0","urllib3==1.25.11"]
```
leikoilja, line 9: `requests` and `urllib3`are already specified in the HA requirements, so we don't need to specify those here. That field is just for dependencies needed specifically for the integration, such as `glocaltokens`

leikoilja: Why do we even need urllib?

leikoilja (changes requested):
Cool! Just a bit more and we're set

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: I will do that

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: but devices from get_google_devices_json() is a list not a dict

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: But devices from .get_google_devices_json() is a list not a dict

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: See above

DurgNomis-drol:
Get_google_devices_json() returns a list, not dict. I know it’s not pretty to loop this way, but maybe we should chance it in glocaltokens

custom_components/ha-glocaltokens/api.py
leikoilja: Uh, sorry, my fault, you are right (https://github.com/leikoilja/glocaltokens/blob/master/glocaltokens/client.py). Although it doesn't make sense the method is called `get_google_devices_json()` but it returns a list 🤔 . I have created a separate issue for that https://github.com/leikoilja/glocaltokens/issues/14

We could still be able to use that kind of iteration
```python
>>> devices = []
>>> devices.append({'deviceName': 'xxx', 'localAuthToken': 'yyy'})
>>> devices.append({'deviceName': 'xxx', 'localAuthToken': 'yyy'})
>>> for device in devices:
...     print(device['deviceName'])
...     print(device['localAuthToken'])
...
xxx
yyy
xxx
yyy

```

custom_components/ha-glocaltokens/api.py
leikoilja: This seems still valid, so i'll unresolve it

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +56,66 @@ def get_google_devices_json(self):
...
+        response = requests.get(url, headers=HEADERS, verify=False, timeout=TIMEOUT)
+
+        return response
+
+    def get_google_devices_information(self):
+
+        devices = self._client.get_google_devices_json()
+
+        for device in range(len(devices)):
+            device_array = devices[1]
```
leikoilja, line 97: We should try making it a bit cleaner if possible :)

leikoilja (changes requested):
Please see comment in the discussions above.

Since this PR will break a current implementation and is still blocked by `glocaltokens` IP autodiscovery, I ll create a new branch `dev` and move this PR to be compared to that. So once it's merged we can keep working on it before merging to `master` and releasing

DurgNomis-drol:
You still have to change the IP manually to make it work. Waiting for device discovery being merged into glocaltokens. But it should be a simpel change when thats done

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +54,66 @@ def get_google_devices_json(self):
...
+        alarm[DATE_TIME] = datetime.utcfromtimestamp(timestamp).strftime(SHOW_DATE_TIMEZONE)
+        alarm[LOCAL_TIME] = datetime.fromtimestamp(timestamp).strftime(SHOW_DATE_AND_TIME)
+        return alarm
+
+    def create_url(self, ip, api_endpoint):
+        url = 'https://{ip}:{port}/{endpoint}'.format(
+            ip=ip,
+            port=str(PORT),
+            endpoint=api_endpoint
+        )
```
leikoilja, line 77: Looks great 💥

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +54,66 @@ def get_google_devices_json(self):
...
+        return url
+
+    def get_alarms_and_timers_from(self, ip, token, endpoint):
+
+        url = self.create_url(ip, endpoint)
+
+        HEADERS[HEADER_CAST_LOCAL_AUTH] = token
+
+        response = requests.get(url, headers=HEADERS, verify=False, timeout=TIMEOUT)
```
leikoilja, line 86: Can we please add here a little comment saying that `verify=False` is needed to avoid SSL security checks? It would be a good reminder for someone looking at the code

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +54,66 @@ def get_google_devices_json(self):
...
+
+        return response
+
+    def get_google_devices_information(self):
+
+        devices = self._client.get_google_devices_json()
```
leikoilja, line 92: Just a heads-up, this method is soon being renamed in `glocaltokens`

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +54,66 @@ def get_google_devices_json(self):
...
+                _LOGGER.error("For device {device} - {error}".format(device=device[DEVICE_NAME], error=API_RETURNED_UNKNOWN))
+                continue
+
+            timers = result[TIMERS]
+            device[TIMERS] = [self.format_timer_information(timer) for timer in timers]
+
+            alarms = result[ALARMS]
+            device[ALARMS] = [self.format_alarm_information(alarm) for alarm in alarms]
+
+        _LOGGER.error(devices)
```
leikoilja, line 118: You probably want to have `debug` logger level here, not `error` 😄

custom_components/ha-glocaltokens/sensor.py
```diff
@@ -21,25 +29,60 @@ class GlocaltokensSensor(GlocaltokensEntity):
...
+    """Representation of a Sensor."""
+
+    def __init__(self, coordinator, entry, device):
+        """Initialize the sensor."""
+        super().__init__(coordinator, entry)
+        self._name = device[DEVICE_NAME]
+        self._alarms = device[ALARMS]
+
+        if len(self._alarms) > 0:
+            self._state = convert_from_ms_to_s(self._alarms[0][FIRE_TIME])
```
leikoilja, line 53: I am not really sure about this one. So we are saying that if `device[ALARMS]` has more than one alarm set then we only take the first one from the list to represent in this sensor? How about the others? I believe this sensor is not really "Google Home alarm sensor" but more like "Google Home next alarm sensor", right?
We could either make it simply like that. But it would be great if we could support multiple alarms for the same google home device in the future, don't you think so?

custom_components/ha-glocaltokens/sensor.py
```diff
@@ -21,25 +29,60 @@ class GlocaltokensSensor(GlocaltokensEntity):
...
+    def __init__(self, coordinator, entry, device):
+        """Initialize the sensor."""
+        super().__init__(coordinator, entry)
+        self._name = device[DEVICE_NAME]
+        self._timers = device[TIMERS]
+
+        if len(self._timers) > 0:
+            self._state = STATE_ON
+        else:
+            self._state = STATE_OFF
```
leikoilja, line 79: Also very similar question (as above) raises here. So if Google Home devices has more then one timer we say that this sensor is on?

custom_components/ha-glocaltokens/utils.py
```diff
@@ -0,0 +1,5 @@
+"""utils.py for google home integration"""
+
+
+def convert_from_ms_to_s(timestamp):
+    return timestamp / 1000
```
leikoilja, line 5: That's great! Having utils file is convenient. I think we should round this number to the closest integer or if it's float then to something with a limited about of digits after the comma. 
What I am afraid of ending up with `convert_from_ms_to_s(...)` return something like `3.781637163` seconds. We can either round it up to the closest integer like `4` or just leave two decimals after the comma and return `3.78`. What do you think?

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +54,66 @@ def get_google_devices_json(self):
...
+
+            if TIMERS not in result and ALARMS not in result:
+                _LOGGER.error("For device {device} - {error}".format(device=device[DEVICE_NAME], error=API_RETURNED_UNKNOWN))
+                continue
+
+            timers = result[TIMERS]
+            device[TIMERS] = [self.format_timer_information(timer) for timer in timers]
+
+            alarms = result[ALARMS]
+            device[ALARMS] = [self.format_alarm_information(alarm) for alarm in alarms]
```
leikoilja, line 116: To be honest I think we need to move `format_alarm_information` into `sensors` file. Now we seem to repeat ourselves. We format the timestamp here, in `api` and then in sensors we call `self._state = convert_from_ms_to_s(self._alarms[0][FIRE_TIME])` again that does the same conversion/formatting. Unless I am missing something.

I propose we keep `api` very blunt. Like:
- we make a request
- we receive data
- we store data (as we have received it) 

And in `sensors` file, we take that received data (as it is) and format it in the way we want it to be displayed/used around Home Assistant.
That way we decouple APIs from our logic. I think it would be a bit of a better approach, because it will allow us to go back into `sensors` any time and change the formating or manipulate received data as we want, without the need to look at `api` file, because there we would simply receive data as is and store it as is, without changing it.

leikoilja (changes requested):
This is becoming so sleek and sexy! Just a few small comments before we can merge it!

Now that we will soon merge these changes into `dev` branch, can you please open a few news issues of what is need to be revisiting in here, like this:
- `[dev] Implement IP address autodiscovery` (we already a ticket like that, so we could just edit it by changing the title to add a prefix `[dev]` to indicate that this change need to be done in `dev` branch. Feel free to edit the description of the ticket to add the reference to the file and lines where the IP is hardcoded now.
- `[dev] Implement timers/alarms updates` (maybe a few lines in the description to explain as you have mentioned in the `EDIT` above) 

And the last one about hammering the google Oauth we'll keep in mind, but `glocaltokens` should take care of it

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: Thanks, im learning so much with project :smile:

custom_components/ha-glocaltokens/sensor.py
DurgNomis-drol: All the alarms are listed as attributes. 

This ways was made so that i could debug it, but my original intention with this was to make similar to how the next alarm sensor function with mobile app.

I think it should always show the next alarm when there is one and then list the rest as attributes. Or we could just create a new alarm sensor for each?

custom_components/ha-glocaltokens/sensor.py
DurgNomis-drol: All timers are listed as attributes, but see my answer above, maybe same things apply to this

custom_components/ha-glocaltokens/utils.py
DurgNomis-drol: because it's unix timstamp i think we should avoid having decimals

custom_components/ha-glocaltokens/sensor.py
leikoilja: I really like your idea about "the next alarm". But it would be hard to implement it that way i guess.
With google home, we can set two types of alarms: single and recurring. Selecting the first alarm from the list doesn't really mean it will be the next alarm. Imagine the list with two alarms where the first on is alarm for every day at 7am, while the second alarm is just at 5pm tonight. And if the now time is 1pm, the next alarm is not going to be the first element in the list, but the second. 

I am not sure what's the best way for us to ultimately go with that but what if we end up implementing generic "GlocaltokenAlarm" for each of the alarms for each of the devices? And inside of the alarm, we can have attributes like: `recurring` and `reccurring_days`. For recurring alarms they will be `recurring = True; recurring_days = "..."` and for single alarms `recuring = False; recurring_days = None`. And then on top of it all we can have a `service` that will be called something like `get_next_alarm`, with input argument of google home's name. Which will go through all Alarm sensors in that particular Google Home's alarms and find the next alarm by comparing them all? 

But even if so, let's just keep it as you have already implemented and we can take that discussion as a whole separate ticket afterwards :)

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: The reason i did it that way was to preserve the original attributes from google home, yes its been called twice, but different things are done each time.

I think your are right about making the API only fetch data and not processing it. I have fixed it now

custom_components/ha-glocaltokens/sensor.py
DurgNomis-drol: First part. Your are absolut right, it was actually what i ment in my first reply, but i did't do a good job of explaining it.
We should sort through all alarms and the choose the one with closes fire_time to now().

The problem with a sensor for each alarm/sensor is that you dont know what the entity name is in HA when you try to call it in an automation or make a card (which i have done). I will create a separate issue for this, so we can find a good way to implement it. :smile:

leikoilja: Oh nice work with the card! 💥  Do you feel like hosting it separately or under the same project?

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +48,48 @@ def get_google_devices_json(self):
...
+        return response
+
+    def get_google_devices_information(self):
+
+        devices = self._client.get_google_devices_json()
+
+        for device in devices:
+            # To avoid keyerror's
+            device[TIMERS] = []
+            device[ALARMS] = []
```
leikoilja, line 74: It just a matter of preference, but you can simply say:
```
times = []
alarms = []
...
response = self.get_alarms_and_timers_from(ip, port, token, API_ENDPOINT_ALARMS)
result = respone.json()
timers = result.get(TIMERS)
alarms = result.get(ALARMS)

if not timers and not alarms:
  _LOGGER.error(...)

device.update({
  TIMERS: timers,
  ALARMS: alarms
})

```

custom_components/ha-glocaltokens/api.py
```diff
@@ -46,3 +48,48 @@ def get_google_devices_json(self):
...
+                _LOGGER.error("For device {device} - API returned {error}".format(device=device[DEVICE_NAME], error=response.status_code))
+                continue
+
+            result = response.json()
+
+            if TIMERS not in result and ALARMS not in result:
+                _LOGGER.error("For device {device} - {error}".format(device=device[DEVICE_NAME], error=API_RETURNED_UNKNOWN))
+                continue
+
+            device.update(result)
```
leikoilja, line 91: Sorry for the confusion on this one. If you are already checking that `results` have TIMERS and ALARMS in that, then it means that at this point you will have them available. You might as well split it in this part: `device.update({TIMERS: [...], ALARMS: [...]})`

custom_components/ha-glocaltokens/sensor.py
```diff
@@ -21,25 +29,60 @@ class GlocaltokensSensor(GlocaltokensEntity):
...
+    """Representation of a Sensor."""
+
+    def __init__(self, coordinator, entry, device):
+        """Initialize the sensor."""
+        super().__init__(coordinator, entry)
+        self._name = device[DEVICE_NAME]
+        self._alarms = [format_alarm_information(alarm) for alarm in device[ALARMS]]
+
+        if len(self._alarms) > 0:
+            self._state = self._alarms[0][FIRE_TIME_IN_S]
```
leikoilja, line 53: You need to import `FIRE_TIME_IN_S`

leikoilja (approved):
Love it! Amazing job, man! 

Just a few last remarks. I have approved it, so feel free the fix the above if you agree with them. Also, please check it works locally for you and then feel free to merge 💥 🚀

DurgNomis-drol:
All set and done, i have opened some issues with what we have discussed above.

This is a big learning experience for me :smile:

custom_components/ha-glocaltokens/api.py
```diff
@@ -52,43 +52,46 @@ def get_android_id(self):
...
+    def get_alarms_and_timers_from(self, device, endpoint):
+        url = self.create_url(device[DEVICE_IP], device[DEVICE_PORT], endpoint)
+        _LOGGER.debug("For device {device} - {url}".format(device=device[DEVICE_NAME], url=url))
+        HEADERS[HEADER_CAST_LOCAL_AUTH] = device[TOKEN]
         response = requests.get(url, headers=HEADERS, verify=False, timeout=TIMEOUT) # verify=False is need to avoid SSL security checks. Othervise it will fail to connect
 
-        return response
+        if response.status_code != HTTP_OK:
+            _LOGGER.error("For device {device} - API returned {error}".format(device=device[DEVICE_NAME], error=response.status_code))
+            return False
```
leikoilja, line 68: Just advice, you could just say `return` here, so your check would look prettier where you call for `get_alarms_and_timers_from`

custom_components/ha-glocaltokens/api.py
```diff
@@ -52,43 +52,46 @@ def get_android_id(self):
...
-                continue
-
-            device.update(result)
+            timers = []
+            alarms = []
+
+            device[DEVICE_IP] = '192.168.0.205' # For testing purpose only
+            device[DEVICE_PORT] = PORT # For testing purpose only
+            result = self.get_alarms_and_timers_from(device, API_ENDPOINT_ALARMS)
+            if result != False:
```
leikoilja, line 84: Then this could just look like: `if result:`
Very pythonic 😄

DurgNomis-drol: Python amazes me every time by how simple and easy it is, i will resolves both of these now

leikoilja:
I am really happy to hear that you are finding this experience useful and are learning from it 💥

PR #21, opened 2021-02-12
Adding alpine devcontainer and documentation

leikoilja:


DurgNomis-drol (approved):
Looks good. This should make it possible to build both grpcio and grpcio-tools.

leikoilja:
You are right, @DurgNomis-drol. All it does is allows to build wheels for development container and provides instructions (that one need to install `apk add gcc g++ linux-headers`) for alpine based HA. You mention that "Homeassistant OS is based on buildroot", where did you find that? I am thinking of trying to open an issue in HA repositories devoted to build images for HA to ask them to install these compilers

DurgNomis-drol:
[Here](https://github.com/home-assistant/operating-system) on the readme in github  for Homeassistant OS

Good ide, it should be a simple fix to add compilers
[Here](https://github.com/home-assistant/docker/blob/master/Dockerfile)

leikoilja:
Those are some great references, thanks, @DurgNomis-drol. Do you by chance know if we need all three libraries `gcc g++ linux-headers` to compline `grpcio`?

DurgNomis-drol:
That thread seems to be for an older version of `grpcio` and `grpcio-tools`.

PR #27, opened 2021-02-14
Update api.py

DurgNomis-drol:
Changed:
`return False` to `return`
and
`if result != False:` to ìf result:`


leikoilja (approved):
Awesome stuff! 💯

PR #34, opened 2021-02-17
Code cleanup

DurgNomis-drol:
Just a little bit of cleanup, so we have a nice base to work from.

### DIFF
--- a/custom_components/ha-glocaltokens/api.py
+++ b/custom_components/ha-glocaltokens/api.py
@@ -12,10 +12,9 @@ from .const import API_RETURNED_UNKNOWN
 from .const import HEADER_CAST_LOCAL_AUTH
 from .const import HEADERS
 from .const import PORT
+from .const import TIMEOUT
 from .const import TIMERS
 from .exceptions import InvalidMasterToken
-
-TIMEOUT = 10
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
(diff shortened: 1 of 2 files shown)


custom_components/ha-glocaltokens/api.py
```diff
@@ -54,37 +53,42 @@ def get_android_id(self):
         return self._client.get_android_id()
 
     def create_url(self, ip, port, api_endpoint):
+        """Creates url to endpoint.
+        Note: port argument is unused because all request must be done to 8443"""
```
leikoilja, line 57: Great comment!

custom_components/ha-glocaltokens/api.py
```diff
@@ -54,37 +53,42 @@ def get_android_id(self):
...
         )
         return url
 
     def get_alarms_and_timers_from(self, device, endpoint):
+        """Fetches timers and alarms from google device"""
         url = self.create_url(device.ip, device.port, endpoint)
         _LOGGER.debug(
             "For device {device} - {url}".format(device=device.device_name, url=url)
         )
+        # verify=False is need to avoid SSL security checks. Othervise it will fail to connect"""
```
leikoilja, line 69: I think we misspelled "othervise" 😄

custom_components/ha-glocaltokens/api.py
```diff
@@ -54,37 +53,42 @@ def get_android_id(self):
...
+        response = requests.get(url, headers=HEADERS, verify=False, timeout=TIMEOUT)
 
         if response.status_code != HTTP_OK:
             _LOGGER.error(
-                "For device {device} - API returned {error}".format(
-                    device=device.device_name, error=response.status_code
+                "For device {device} - API returned {error}: {reason}".format(
+                    device=device.device_name,
+                    error=response.status_code,
+                    reason=response.reason,
```
leikoilja, line 78: Are you sure the error message lands in `response.reason`? Some other http requests i have worked with were returning error message in `response.message`, but i m not sure what would be here.

custom_components/ha-glocaltokens/const.py
```diff
@@ -31,18 +30,21 @@
...
-# API
+# DEVICE PORT
 PORT = 8443
+
+# API
 API_ENDPOINT_ALARMS = "setup/assistant/alarms"
 HEADER_CAST_LOCAL_AUTH = "cast-local-authorization-token"
 HEADER_CONTENT_TYPE = "content-type"
-
+# HEADERS
```
leikoilja, line 40: Maybe just an extra empty line above would help being consistent?

custom_components/ha-glocaltokens/const.py
```diff
@@ -55,17 +57,14 @@
...
 
 # Access token only lives about 1 hour
 # Update often to fetch timers in timely manner
 SCAN_INTERVAL = 15  # Seconds
 
-DEVICE_NAME = "deviceName"
-DEVICE_IP = "ip"
-DEVICE_PORT = "port"
-TOKEN = "localAuthToken"
+# JSON name when retrieving information fra devices
```
leikoilja, line 67: Maybe something like "json parameter values when retrieving information from devices"?

leikoilja (approved):
Great cleanup, thank you! 💯 🚀 

I will approve, feel free to either fix/skip the comments I have added, but once fixed - let's merge it in 💥

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: :facepalm:

custom_components/ha-glocaltokens/api.py
DurgNomis-drol: As far as i can see requests lib does not return message but instead reason if it fails [requests-lib](https://requests.readthedocs.io/en/master/api/#requests.Response)

We can also change it to `response.text` instead?

DurgNomis-drol: I have changed it to `response.text` because it will return the most amount of info in case of error

leikoilja:
Superb, thanks 🚀

PR #37, opened 2021-02-18
Disabling zeroconf from glocaltokens

leikoilja:
- Fetching list of devices, not json.
- Disabling zeroconf discovery from glocaltokens

Note: it depends on `master` branch of `glocaltokens`, which is not yet released. `disable_discovery` is only coming in a next release

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,7 @@
 name: Manage labels
 
 on:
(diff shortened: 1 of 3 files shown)


DurgNomis-drol (approved):

leikoilja:
Nice! We ll merge it once 'glocaltokens' gets a new release update 💥🤩

PR #38, opened 2021-02-18
Added warning if one or more devices are offline

DurgNomis-drol:
Sorry for the typo. I'm not good friends with PyCharm, so have reverted to using command line


DurgNomis-drol:
We should probably expand on this in the future to differentiate between 404 or because of a bad IP

leikoilja:
Oh, flip. I have just rebased `dev` branch to master and that messed up your PR a bit. Sorry for that. 
Can you take a look to fix merge conflicts we have here now, or do you want me to take a look at it? 😄

DurgNomis-drol:
I will give be it a try :smile:

I should rebase my local fork from this dev branch right?

DurgNomis-drol:
I think i fixed it now :smile: Can you please confirm? @leikoilja

custom_components/ha-google-home/api.py
```diff
@@ -52,18 +52,29 @@ def get_android_id(self):
...
-            ip=ip, port=str(PORT), endpoint=api_endpoint
+            ip=ip, port=str(port), endpoint=api_endpoint
         )
         return url
 
     def get_alarms_and_timers_from(self, device, endpoint):
         """Fetches timers and alarms from google device"""
-        url = self.create_url(device.ip, device.port, endpoint)
+        if device.ip is None:
+            return
```
leikoilja, line 76: Can we do that check outside this method? It would make a bit more sense 
See comment below

custom_components/ha-google-home/api.py
```diff
@@ -104,10 +116,26 @@ def get_google_devices_information(self):
                             device=device.device_name, error=API_RETURNED_UNKNOWN
                         )
                     )
+            else:
+                offline_devices.append(device.device_name)
```
leikoilja, line 120: I think we should add the ip check inside of the `for` loop 
Sort of 
```
for device in devices:
...
if device.ip:
  timers = []
  alarms = []
  result = ...
else:
  offline_devices.append(...)
```

custom_components/ha-google-home/api.py
```diff
@@ -104,10 +116,26 @@ def get_google_devices_information(self):
...
 
             device.timers = timers
             device.alarms = alarms
             _LOGGER.debug(device)
 
+        """Gives the user a warning if the device is offline, but will not remove entities or device from
+        HA device registry"""
```
leikoilja, line 127: The comment inside of method rarely is done as docstring `""" ... """`, even if it's a multiline comment, let's just keep it as a comment `# ....`

custom_components/ha-google-home/api.py
```diff
@@ -104,10 +116,26 @@ def get_google_devices_information(self):
...
+        if len(offline_devices) > 0:
```
leikoilja, line 128: No need to check for `len` explicitly. Simply `if offline_devices:` will do the same

custom_components/ha-google-home/api.py
```diff
@@ -104,10 +116,26 @@ def get_google_devices_information(self):
...
+                "A device is " if len(offline_devices) == 1 else "Some devices are "
+            )
+            _LOGGER.warning(
+                "{one_or_more}offline, so no information could be retrieved, will try again later - {devices}".format(
+                    one_or_more=if_one,
+                    devices=self.format_offline_devices_to_human_string(
+                        offline_devices
+                    ),
+                )
+            )
```
leikoilja, line 139: Instead of `offline_devices` list conversion to `str`, which is a bit complex, I propose we do a`for` loop and print a new warning per each device, just informing that the device is offline. I think it will take down a lot of unnecessary complexity away :)

leikoilja (changes requested):
Amazing stuff! 🚀  Just a few small remarks

custom_components/ha-google-home/api.py
DurgNomis-drol: The reason I did it this way was because that sometime the request to the device will fail for no good reason, so to not overflow the user with warnings, i thought printing it on one line instead. I'm also working on a way to check if an IP is valid or not.

custom_components/ha-google-home/api.py
DurgNomis-drol: I will moved it

custom_components/ha-google-home/api.py
DurgNomis-drol: After thinking about it, your right, i have changed :smile:

custom_components/ha-google-home/api.py
```diff
@@ -52,62 +54,88 @@ def get_android_id(self):
...
     @staticmethod
     def create_url(ip, port, api_endpoint):
-        """Creates url to endpoint.
+        """First it checks if a valid IP have been provided and then it creates a url to endpoint.
         Note: port argument is unused because all request must be done to 8443"""
-        url = "https://{ip}:{port}/{endpoint}".format(
-            ip=ip, port=str(PORT), endpoint=api_endpoint
-        )
-        return url
+        if re.search(IP_CHECK_REGEX, ip):
```
leikoilja, line 71: Let's better use inbuild `glocaltokens` check that is more robust. Check it [here](https://github.com/leikoilja/glocaltokens/blob/master/glocaltokens/utils/network.py)

`from glocaltokens.utils.network import is_valid_ipv4_address`
It will return True/False

custom_components/ha-google-home/api.py
```diff
@@ -52,62 +54,88 @@ def get_android_id(self):
...
-                timers = result.get(TIMERS)
-                alarms = result.get(ALARMS)
-
-                if not timers and not alarms:
-                    _LOGGER.error(
-                        "For device {device} - {error}".format(
-                            device=device.device_name, error=API_RETURNED_UNKNOWN
+            if device.ip:
+                result = self.get_alarms_and_timers_from(device, API_ENDPOINT_ALARMS)
+                if result:
```
leikoilja, line 114: What if this returns nothing or a wrong data?
I suggest remove this check, because the next check `if not timers and not alarms` will cater for it and log error

custom_components/ha-google-home/api.py
```diff
@@ -52,62 +54,88 @@ def get_android_id(self):
...
+                    timers = result.get(TIMERS)
+                    alarms = result.get(ALARMS)
+
+                    if not timers and not alarms:
+                        _LOGGER.error(
+                            "For device {device} - {error}".format(
```
leikoilja, line 120: Shall we make the error message more elaborate "Failed to fetch timers/alarms information from device {device} ..."?

custom_components/ha-google-home/api.py
```diff
@@ -52,62 +54,88 @@ def get_android_id(self):
...
+                                device=device.device_name, error=API_RETURNED_UNKNOWN
+                            )
                         )
-                    )
+            else:
+                offline_devices.append(device.device_name)
 
             device.timers = timers
             device.alarms = alarms
             _LOGGER.debug(device)
```
leikoilja, line 130: This should be moved inside the `if device.ip`

custom_components/ha-google-home/api.py
```diff
@@ -52,62 +54,88 @@ def get_android_id(self):
         """Generate random android_id"""
         return self._client.get_android_id()
 
+    @staticmethod
+    def format_offline_devices_to_human_string(device_list):
```
leikoilja, line 58: This is no longer needed, let's remove it?

leikoilja (changes requested):
A few small comments

custom_components/ha-google-home/api.py
DurgNomis-drol: If this is moved it will unregistered all the devices from HA every time it goes offline

custom_components/ha-google-home/api.py
DurgNomis-drol: It should not be anything that causes concern to the user, this will happen regularly because the request sometimes fails for no good reason. But it could probably be worded better then what i have done. But i think that goes for all error, debug and warning messages so far :smile:

custom_components/ha-google-home/api.py
DurgNomis-drol: I'm not use i understand what you are referering to?

custom_components/ha-google-home/api.py
DurgNomis-drol: okay

custom_components/ha-google-home/api.py
leikoilja: Sorry, i was unclear. I was confused what if `result` is nothing. Then we just skip the device and do nothing. I suggest we remove `if result:` at all and rely on later check `if not timers and not alarms`

custom_components/ha-google-home/api.py
leikoilja: Hm, i m a bit confused, but sure. Will need to dive deeper into that later when i can get an access to my setup :) Let's just keep it then

custom_components/ha-google-home/api.py
DurgNomis-drol: `result` will return `False` if a http error occurred, and `if not timers and not alarms` is just to check if the JSON returned includes the necessary attributes.

`if not timers and not alarms` error should only happen if Google changes the API

custom_components/ha-google-home/api.py
DurgNomis-drol: Actually, your right, we can move it. I was the one getting confused :laughing:

custom_components/ha-google-home/api.py
DurgNomis-drol: Actually, shouldn't we use exceptions to make error messages?

leikoilja: It didn't seem from the documentation that HA wants custom integrations to raise exceptions, so the logger is a good way to go about it :)

custom_components/ha-google-home/api.py
leikoilja: oh, you are right, sorry, forgot that `results` is a json response. Maybe we could also rename it to `result_json` for better clarity? 😄 Will just mark it resolved then 👍

DurgNomis-drol: Good idea :smile:

custom_components/ha-google-home/api.py
```diff
@@ -86,28 +86,42 @@ def get_alarms_and_timers_from(self, device, endpoint):
...
+
+                    if not timers and not alarms:
+                        _LOGGER.error(
+                            "For device {device} - {error}".format(
+                                device=device.device_name, error=API_RETURNED_UNKNOWN
+                            )
                         )
-                    )
+            else:
+                offline_devices.append(device.device_name)
```
leikoilja, line 111: You call the list `offline_devices` but only append their names. You can just append `device` object

custom_components/ha-google-home/api.py
```diff
@@ -86,28 +86,42 @@ def get_alarms_and_timers_from(self, device, endpoint):
...
 
+        # Gives the user a warning if the device is offline, but will not remove entities or device from
+        # HA device registry
+        if offline_devices:
+            for device in offline_devices:
+                _LOGGER.warning(
+                    "Failed to fetch timers/alarms information from device {device}. Will try again later.".format(
+                        device=device
+                    )
+                )
```
leikoilja, line 125: Since you have `device` objects inside your list, just print the name here, if you wish
```
LOGGER.warning(
"Failed to fetch timers/alarms information from device {device_name}. Will try again later.".format(device_name=device.device_name)
)
```

leikoilja (approved):
Minor stuff and merge it in! 💥  🚀

PR #42, opened 2021-02-21
Renaming classes to match Google-Home

leikoilja:
Simply renaming all other places from any `glocaltokens` reference to `google-home`. 
If anyone has better suggestions for class naming - I'm open :)


DurgNomis-drol (approved):
Awesome! :rocket:

PR #46, opened 2021-03-03
Removed `@gmail.com` omission indication.

### LINKED ISSUE #45: Relax requirements for account name
**Is your feature request related to a problem? Please describe.**
Requiring stripping of `@gmail.com` is too restrictive. Also note that Google accounts can be on custom domains with G Suite and they probably also should be supported.

**Describe the solution you'd like**
This can be done programatically.

ArnyminerZ:
Closes #45


leikoilja (changes requested):
Please see `config_flow.py` file where we create `username` variable by appending `@gmail.com`. We should probs remove it

ArnyminerZ (commented):
Removed appending of `@gmail.com`

leikoilja (approved):
Let's merge it! Thanks a lot!

PR #47, opened 2021-03-03
Making master stable again! 💥

leikoilja:
This is rather big PR that is aimed to refactor and fix `master` branch to a minimum working version.

P.S. Everything works on this branch, except the update of data. It fetches the data periodically RIGHT, but does not yet update the sensors with the new data. It's something to do with coordinator that we need to attach a signal to that i guess. I just didn't have time for it right now, will fix it in the next few days. Apart from that, when restarting HA it should fetch all data and initialise sensors correctly.

### DIFF
--- a/custom_components/ha-google-home/__init__.py
+++ b/custom_components/ha-google-home/__init__.py
@@ -8,14 +8,13 @@ import asyncio
 import logging
 from datetime import timedelta
 
-from async_timeout import timeout
+from homeassistant.components import zeroconf
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import Config
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator
-from homeassistant.helpers.update_coordinator import UpdateFailed
 
 from .api import GlocaltokensApiClient
 from .const import CONF_ANDROID_ID
@@ -23,10 +22,10 @@ from .const import CONF_PASSWORD
 from .const import CONF_USERNAME
 from .const import DOMAIN
 from .const import PLATFORMS
-from .const import SCAN_INTERVAL
+from .const import SENSOR
 from .const import STARTUP_MESSAGE
+from .const import UPDATE_INTERVAL
 
-SCAN_INTERVAL = timedelta(seconds=SCAN_INTERVAL)
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
@@ -47,9 +46,25 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     android_id = entry.data.get(CONF_ANDROID_ID)
 
     session = async_get_clientsession(hass)
-    client = GlocaltokensApiClient(username, password, session, android_id)
 
-    coordinator = GoogleHomeDataUpdateCoordinator(hass, client=client)
+    glocaltokens_client = GlocaltokensApiClient(
+        hass, username, password, session, android_id
+    )
+    zeroconf_instance = await zeroconf.async_get_instance(hass)
+
+    async def async_update_data():
+        return await glocaltokens_client.get_google_devices_information(
+            zeroconf_instance
+        )
+
+    coordinator = DataUpdateCoordinator(
+        hass,
+        _LOGGER,
+        name=SENSOR,
+        update_method=async_update_data,
+        update_interval=timedelta(seconds=UPDATE_INTERVAL),
+    )
+
     await coordinator.async_refresh()
 
     if not coordinator.last_update_success:
@@ -57,53 +72,23 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 
     hass.data[DOMAIN][entry.entry_id] = coordinator
 
+    # Offload the loading of entities to the platform
     for platform in PLATFORMS:
-        if entry.options.get(platform, True):
-            coordinator.platforms.append(platform)
-            hass.async_add_job(
-                hass.config_entries.async_forward_entry_setup(entry, platform)
-            )
+        hass.async_add_job(
+            hass.config_entries.async_forward_entry_setup(entry, platform)
+        )
 
     entry.add_update_listener(async_reload_entry)
     return True
 
 
-class GoogleHomeDataUpdateCoordinator(DataUpdateCoordinator):
-    """Class to manage fetching data from the API."""
-
-    def __init__(
-        self,
-        hass: HomeAssistant,
-        client: GlocaltokensApiClient,
-    ) -> None:
-        """Initialize."""
-        self.api = client
-        self.platforms = []
-
-        super().__init__(hass, _LOGGER, name=DOMAIN, update_interval=SCAN_INTERVAL)
-
-    def _update_data(self) -> dict:
-        """Fetch local auth tokens from Google via sync functions."""
-        return self.api.get_google_devices_information()
-
-    async def _async_update_data(self):
-        """Update data via library."""
-        try:
-            async with timeout(10):
-                return await self.hass.async_add_executor_job(self._update_data)
-        except Exception as exception:
-            raise UpdateFailed(f"Invalid response from API: {exception}") from exception
-
-
 async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     """Handle removal of an entry."""
-    coordinator = hass.data[DOMAIN][entry.entry_id]
     unloaded = all(
         await asyncio.gather(
             *[
                 hass.config_entries.async_forward_entry_unload(entry, platform)
                 for platform in PLATFORMS
-                if platform in coordinator.platforms
             ]
         )
     )
--- a/custom_components/ha-google-home/__init__.py
+++ b/custom_components/ha-google-home/__init__.py
@@ -56,7 +56,9 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         hass,
         _LOGGER,
         name=SENSOR,
-        update_method=lambda: glocaltokens_client.get_google_devices_information(zeroconf_instance),
+        update_method=lambda: glocaltokens_client.get_google_devices_information(
+            zeroconf_instance
+        ),
         update_interval=timedelta(seconds=UPDATE_INTERVAL),
     )
 
--- a/custom_components/ha-google-home/__init__.py
+++ b/custom_components/ha-google-home/__init__.py
@@ -45,7 +45,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     password = entry.data.get(CONF_PASSWORD)
     android_id = entry.data.get(CONF_ANDROID_ID)
 
-    session = async_get_clientsession(hass)
+    session = async_get_clientsession(hass, verify_ssl=False)
 
     glocaltokens_client = GlocaltokensApiClient(
         hass, username, password, session, android_id
--- a/custom_components/ha-google-home/__init__.py
+++ b/custom_components/ha-google-home/__init__.py
@@ -50,18 +50,13 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     glocaltokens_client = GlocaltokensApiClient(
         hass, username, password, session, android_id
     )
+
     zeroconf_instance = await zeroconf.async_get_instance(hass)
-
-    async def async_update_data():
-        return await glocaltokens_client.get_google_devices_information(
-            zeroconf_instance
-        )
-
     coordinator = DataUpdateCoordinator(
         hass,
         _LOGGER,
         name=SENSOR,
-        update_method=async_update_data,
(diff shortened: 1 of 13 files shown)


DurgNomis-drol:
This looks awesome, way over my level, but i think it addresses a lot of the issues we have talked about in other threads. 😁

leikoilja:
Thanks! It does address a lot of issues, @DurgNomis-drol. Sorry it's a bit packed.
 It would be much better to split into many smaller PRs, but since our master is in a really bad shape, let's just try fixing it in one BIG move, so we can bring it up to speed as soon as possible. And then we can/should address individual issues in separate PRs.

DurgNomis-drol:
@leikoilja I totally agree. Can't wait to test it out once it is merged.

leikoilja:
Totally! We ll need a bunch of testing when merged 😂🤩💥 I m excited to see where we get with this 🤩💥 i m also wondering if we can find a way to extract reminders as well 🤷‍♂️

DurgNomis-drol:
Same here 💥, and it seems that there are a lot of people interested in it too

That would be really cool and extremely useful to have

ArnyminerZ:
Awesome work! Yeah, I think it's better to move everything in a single PR if `master` is not working.

ArnyminerZ (approved):

leikoilja:
Awesome! Thanks for the review, @ArnyminerZ.
Happy we are all in the same page with that one! I ll try to fix the coordinator data today, but have some plans after work today. Might be tomorrow. Will push the fix here, when i have it and then we can merge for testing :)

custom_components/ha-google-home/__init__.py
```diff
@@ -46,64 +45,50 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+
+    glocaltokens_client = GlocaltokensApiClient(
+        hass, username, password, session, android_id
+    )
+    zeroconf_instance = await zeroconf.async_get_instance(hass)
+
+    async def async_update_data():
+        return await glocaltokens_client.get_google_devices_information(
+            zeroconf_instance
+        )
```
KapJI, line 58: Why not using lamda here? I think it's cleaner than defining sub function in the middle.

leikoilja: You are right! Haven't worked with lambda much before. Do you mind sharing a snippet how you'd do that? :)

KapJI: There are few tricky parts here.
1. There are no async lamdas in python. But you can do
```
async_update_data = lamda: glocaltokens_client.get_google_devices_information(zeroconf_instance)
```
It will return awaitable as data coordinator expects.
2. PEP8 discourages using named lamdas (assigning lamda to variable): https://www.python.org/dev/peps/pep-0008/#programming-recommendations

So there are 2 options here:
1. Inline lamda in `DataCoordinator` constructor.
2. Move `async_update_data()` to the top of `async_setup_entry()`, so it's a bit better separated from the rest of the function code. It's probably worth moving `zeroconf_instance` fetching as well to have it before `async_update_data` definition, but it should work even without this.

custom_components/ha-google-home/sensor.py
```diff
@@ -1,38 +1,53 @@
...
+        if device_data[LABEL_ALARMS]:
+            devices_to_add.append(
+                GoogleHomeAlarmSensor(coordinator, entry, device_name, device_data)
+            )
+        if device_data[LABEL_TIMERS]:
+            devices_to_add.append(
+                GoogleHomeTimerSensor(coordinator, entry, device_name, device_data)
             )
+        if devices_to_add:
+            async_add_devices(devices_to_add)
```
KapJI, line 40: This basically means if there is no alarm or timer when platform is setup (home assistant started) then corresponding device won't be created and it won't get any updates. You should probably revert this.

KapJI (changes requested):

custom_components/ha-google-home/sensor.py
KapJI, line 40: This basicallt means if there is no alarm or timer then corresponding device won't be created during platform setup (home assistant start). And it won't get any updates in the future. You should probably revert this.

KapJI (changes requested):

custom_components/ha-google-home/sensor.py
```diff
@@ -46,11 +61,14 @@ def device_state_attributes(self):
...
     """Representation of a Sensor."""
 
-    def __init__(self, coordinator, entry, device):
+    def __init__(self, coordinator, entry, device_name, device_data):
         """Initialize the sensor."""
         super().__init__(coordinator, entry)
-        self._name = device.device_name
-        self._alarms = [format_alarm_information(alarm) for alarm in device.alarms]
+        self._name = device_name
+        self._alarms_timestamp = device_data[LABEL_ALARMS]
```
KapJI, line 68: This is list of alarms. Each alarm is a dict. You clearly meant something else here. Probably `device_data[LABEL_ALARMS][0]["fire_time"] if len(device_data[LABEL_ALARMS]) > 0 else None`

custom_components/ha-google-home/sensor.py
```diff
@@ -46,11 +61,14 @@ def device_state_attributes(self):
...
+        self._alarms = [
+            format_alarm_information(alarm) for alarm in device_data[LABEL_ALARMS]
```
KapJI, line 70: `format_alarm_information` expects timestamp, so it should be `alarm["fire_time"]`

custom_components/ha-google-home/sensor.py
```diff
@@ -70,11 +91,14 @@ def device_state_attributes(self):
...
+    def __init__(self, coordinator, entry, device_name, device_data):
         """Initialize the sensor."""
         super().__init__(coordinator, entry)
-        self._name = device.device_name
-        self._timers = [format_timer_information(timer) for timer in device.timers]
+        self._name = device_name
+        self._timers_timestamp = device_data[LABEL_TIMERS]
+        self._timers = [
+            format_timer_information(timer) for timer in device_data[LABEL_TIMERS]
+        ]
```
KapJI, line 101: Same here.

custom_components/ha-google-home/sensor.py
```diff
@@ -1,38 +1,53 @@
...
-                ]
+    for device_name, device_data in coordinator.data.items():
+        devices_to_add = []
+        if device_data[LABEL_TOKEN]:
+            devices_to_add.append(
+                GoogleHomeTokenSensor(coordinator, entry, device_name, device_data)
+            )
+        if device_data[LABEL_ALARMS]:
+            devices_to_add.append(
+                GoogleHomeAlarmSensor(coordinator, entry, device_name, device_data)
```
KapJI, line 33: I'd probably pass `device_data[LABEL_ALARMS]` here instead of `device_data`.

KapJI (changes requested):

custom_components/ha-google-home/utils.py
```diff
@@ -16,19 +16,21 @@ def convert_from_ms_to_s(timestamp):
     return round(timestamp / 1000)
 
 
-def format_timer_information(timer):
-    timer[FIRE_TIME_IN_S] = convert_from_ms_to_s(timer[FIRE_TIME])
+def format_timer_information(timer_timestamp):
```
KapJI, line 19: `timer_timestamp` is very confusing name. This is a whole dict with timer attributes. I'd keep old name.

custom_components/ha-google-home/utils.py
```diff
@@ -16,19 +16,21 @@ def convert_from_ms_to_s(timestamp):
...
 
-    duration = convert_from_ms_to_s(timer[ORIGINAL_DURATION])
+    duration = convert_from_ms_to_s(timer_timestamp[ORIGINAL_DURATION])
     timer[DURATION] = datetime.utcfromtimestamp(duration).strftime(SHOW_TIME_ONLY)
     return timer
 
 
-def format_alarm_information(alarm):
-    alarm[FIRE_TIME_IN_S] = convert_from_ms_to_s(alarm[FIRE_TIME])
+def format_alarm_information(alarm_timestamp):
```
KapJI, line 31: Same here, `alarm_timestamp` is a dict.

KapJI (changes requested):

custom_components/ha-google-home/__init__.py
KapJI: Actually I see this is quite a common practice in home assistant components to use `async def async_update_data()`, probably it's better to keep it this way, may only move a bit.

custom_components/ha-google-home/sensor.py
```diff
@@ -60,8 +78,11 @@ def __init__(self, coordinator, entry, device):
     @property
     def device_state_attributes(self):
         """Return the state attributes."""
+        _LOGGER.debug("ALARM ATTRIBUTE data", self.coordinator.data)
```
KapJI, line 81: You're not seeing any data here because logger supports only 1 argument unlike `print`. It should be: `_LOGGER.debug(f"ALARM ATTRIBUTE data {self.coordinator.data}")

KapJI: It's not very precise, actually logger supports format string and extra arguments: https://docs.python.org/3/library/logging.html#logging.debug
But anyway `coordinator.data` is there, you only need to process it.

custom_components/ha-google-home/__init__.py
leikoilja: Awesome, thanks! 💥

KapJI:
You can do something like this to fix alarms updating: https://pastebin.com/hg9z12WG

leikoilja:
This is exciting, thanks a lot, @KapJI, for the great feedback and suggestions! It all seems to work now! 💥 

The only thing I don't really like is how we fetch "state" and "state_attributes". I believe it updates the data correctly but is not updating "last updated" variable correctly, so from developer tools it looks like the sensors are not being updated when in reality their data is. It should be a minor refactoring, i believe that implementing "update" method on the sensor is the right way to go to update both sensor data and "last updated" state variable. But will leave it for another refactoring PR. 😄

custom_components/ha-google-home/sensor.py
```diff
@@ -46,46 +60,62 @@ def device_state_attributes(self):
...
 
-        if len(self._timers) > 0:
-            self._state = STATE_ON
-        else:
-            self._state = STATE_OFF
+    @property
+    def state(self):
+        timers = self.get_timers_data()
+        state = timers[0][TIME_LEFT] if timers else STATE_OFF
+        return state
```
KapJI, line 104: nit:
```suggestion
        return timers[0][TIME_LEFT] if timers else STATE_OFF
```

custom_components/ha-google-home/sensor.py
```diff
@@ -46,46 +60,62 @@ def device_state_attributes(self):
...
 class GoogleHomeTimerSensor(GoogleHomeTimersEntity):
     """Representation of a Sensor."""
 
-    def __init__(self, coordinator, entry, device):
+    def __init__(self, coordinator, entry, device_name, timers):
         """Initialize the sensor."""
         super().__init__(coordinator, entry)
-        self._name = device.device_name
-        self._timers = [format_timer_information(timer) for timer in device.timers]
+        self._name = device_name
```
KapJI, line 98: In fact `timers` is never used (`pylint` or `flake8` should warn about this, probably worth setting them up) and you can remove it completely. I believe `entry` is also not used in entity, can be stripped off as well.

custom_components/ha-google-home/sensor.py
```diff
@@ -46,46 +60,62 @@ def device_state_attributes(self):
...
     def device_state_attributes(self):
         """Return the state attributes."""
         return {
-            "alarms": self._alarms,
+            "alarms": self.get_alarms_data(),
             "device": str(self.name),
             "integration": DOMAIN,
         }
 
+    def get_alarms_data(self):
```
KapJI, line 83: you can mark it private
```suggestion
    def _get_alarms_data(self):
```

custom_components/ha-google-home/sensor.py
```diff
@@ -1,38 +1,52 @@
...
+_LOGGER: logging.Logger = logging.getLogger(__package__)
+
+
 async def async_setup_entry(hass, entry, async_add_devices):
     """Setup sensor platform."""
     coordinator = hass.data[DOMAIN][entry.entry_id]
-    for device in coordinator.data:
-        if device.local_auth_token:
+    for device_name, device_data in coordinator.data.items():
+        if device_data[LABEL_TOKEN]:
```
KapJI, line 26: yeah, this makes sense

KapJI (approved):
Few minor comments. I think it's good to go.

custom_components/ha-google-home/sensor.py
leikoilja: Hm, weird that `flake8` didn't catch it, we have it installed into `pre-commit` hooks. Maybe that's because I use similarly named variables in the class. Thanks for spotting on those. I will leave it just for now, I believe there must be more work done further on sensor classes to implement proper data updating using `update` class (but I need to do a better lookup if that's true). So will keep these for now, but surely stripping them all down if not needed further down. :)

leikoilja:
Awesome, thanks a lot for the help, @KapJI 💥

PR #55, opened 2021-03-05
Change state and time duration

### LINKED ISSUE #50: Alarm state format
## Problem
In #47 alarm state format was changed from unix timestamp to formatted string which is nice and makes it more readable. But I think it should also include time zone because it's a local time. 

Before March 28 I live in UTC and can't check this but Sweden uses CET and I believe if in template editor in developer tools you try this, these timestampts won't be equal:
```
Converted timestamp: {{ as_timestamp(states("sensor.<google_home_name>_alarms")) }}
Fire time: {{ state_attr("sensor.<google_home_name>_alarms", "alarms")[0]["fire_time"] }}
```
This makes it harder to use state in automations.

## Solution
Add time zone to `TIME_STR_FORMAT`. `%z` should be compatible with ISO datetime format. It should look like `2021-03-04 13:41:04+0100`

### LINKED ISSUE #52: Change attribute time_left to duration instead

Latest master
## Configuration

```yaml
timers:
  - fire_time: 1614879264
    date_time_utc: '2021-03-04 17:34:24'
    local_time: '2021-03-04 18:34:24'
    time_left: '0:30:00'
device: Kitchen timers
integration: ha-google-home
friendly_name: Kitchen timers
icon: 'mdi:timer-sand'
```

## Describe the bug

`time_left` should be `duration` instead. It does not count down, so i think the name is misleading

leikoilja:
- Changing timer and alarm states to on/off correspondingly if there are any values set.
- Adding `local_time_iso` key to timer and alarm sensors for easier automation

Closes #50
Closes #52

### DIFF
--- a/custom_components/ha-google-home/const.py
+++ b/custom_components/ha-google-home/const.py
@@ -53,9 +53,9 @@ TIMEOUT = 10  # Request Timeout in seconds
 
 # TIMERS & ALARMS ATTRIBUTE NAMES
 FIRE_TIME = "fire_time"
-DATETIME_UTC = "date_time_utc"
 LOCAL_TIME = "local_time"
-TIME_LEFT = "time_left"
+LOCAL_TIME_ISO = "local_time_iso"
+DURATION = "duration"
 ORIGINAL_DURATION = "original_duration"
 
 # TIMESTRINGS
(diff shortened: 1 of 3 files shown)


DurgNomis-drol (approved):
Looks good to me 😄

KapJI:
Looks great, thanks!

Issue #61, opened 2021-03-05
401 Unauthorised error handling


leikoilja:
## Describe the bug
We cache local authentication token in `glocaltokens` package, but if the token expires - we don't know and requests fail with 401.
There is a new feature in `glocaltokens` to force reload homegraph that will provide new local auth tokens. See [here](https://github.com/leikoilja/glocaltokens/pull/37). What we need to do is introduce 401 error handling to our `session.get` request that will re-run `get_google_devices` method with `force_homegraph_reload=True`

PR #62, opened 2021-03-06
401 error handling and caching google devices

### LINKED ISSUE #60: Make updates faster
## Problem

Updates now are unreasonably slow and take around 20 seconds.

## Solution
- Run zeroconf only for setup
- Reuse tokens and don't refresh them for every update, they should be valid at least for 1 hour.
- Query alarms and timers from devices in parallel (probably already done)

Basically we can always assume that token is still valid and if request fails, only then we need to fetch the new one and retry the request.

With this update interval can be pretty small like 10 or 15 seconds and there is no need in #58

leikoilja:
- Adding google home device caching on the api class
- Adding error handling when the local auth token expires and the response to google home returns 401 
- Making sensor update rate back 15 sec

Closes #60 and #61 

With these changes, a new fresh homegraph request-response cycle takes about 20 seconds(should only happen 1-2 per day) but all the subsequent api calls to the devices themselves take about 0.2 sec

### DIFF
--- a/custom_components/ha-google-home/__init__.py
+++ b/custom_components/ha-google-home/__init__.py
@@ -56,7 +56,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         hass,
         _LOGGER,
         name=SENSOR,
-        update_method=lambda: glocaltokens_client.update_google_devices_information(),
+        update_method=glocaltokens_client.update_google_devices_information,
         update_interval=timedelta(seconds=UPDATE_INTERVAL),
     )
 
--- a/custom_components/ha-google-home/__init__.py
+++ b/custom_components/ha-google-home/__init__.py
@@ -47,18 +47,16 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 
     session = async_get_clientsession(hass, verify_ssl=False)
 
+    zeroconf_instance = await zeroconf.async_get_instance(hass)
     glocaltokens_client = GlocaltokensApiClient(
-        hass, username, password, session, android_id
+        hass, username, password, session, android_id, zeroconf_instance
     )
 
-    zeroconf_instance = await zeroconf.async_get_instance(hass)
     coordinator = DataUpdateCoordinator(
         hass,
         _LOGGER,
         name=SENSOR,
-        update_method=lambda: glocaltokens_client.get_google_devices_information(
-            zeroconf_instance
-        ),
+        update_method=lambda: glocaltokens_client.update_google_devices_information(),
         update_interval=timedelta(seconds=UPDATE_INTERVAL),
     )
 
--- a/custom_components/ha-google-home/api.py
+++ b/custom_components/ha-google-home/api.py
@@ -1,5 +1,6 @@
 """Sample API Client."""
 import logging
+from asyncio import gather
 
 import aiohttp
 from glocaltokens.client import GLocalAuthenticationTokens
@@ -16,7 +17,6 @@ from .const import JSON_ALARM
 from .const import JSON_TIMER
 from .const import LABEL_ALARMS
 from .const import LABEL_TIMERS
-from .const import LABEL_TOKEN
 from .const import PORT
 from .const import TIMEOUT
 from .exceptions import InvalidMasterToken
@@ -92,8 +92,9 @@ class GlocaltokensApiClient:
         )
         return url
 
-    async def get_alarms_and_timers(self, url, device):
+    async def get_alarms_and_timers(self, device):
         """Fetches timers and alarms from google device"""
+        url = self.create_url(device.ip, PORT, API_ENDPOINT_ALARMS)
         _LOGGER.debug(
             "Fetching data from Google Home device {device} - {url}".format(
                 device=device.device_name, url=url
@@ -115,9 +116,10 @@ class GlocaltokensApiClient:
                 self.google_devices = []
             elif response.status == HTTP_NOT_FOUND:
                 _LOGGER.debug(
-                    "Failed to fetch data from {device}, API returned {code}: The devics is not Google Home compatable and has to alarms/timers".format(
+                    "Failed to fetch data from {device}, API returned {code}: The device(hardware='{hardware}') is not Google Home compatable and has no alarms/timers".format(
                         device=device.device_name,
                         code=response.status,
+                        hardware=device.hardware,
                     )
                 )
             else:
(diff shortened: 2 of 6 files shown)


DurgNomis-drol (approved):
Awesome! Great job 🚀 😄

leikoilja:
@KapJI please take a look if that satisfies the desired behavior of #60  :)

custom_components/ha-google-home/__init__.py
```diff
@@ -47,18 +47,16 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
 
-    zeroconf_instance = await zeroconf.async_get_instance(hass)
     coordinator = DataUpdateCoordinator(
         hass,
         _LOGGER,
         name=SENSOR,
-        update_method=lambda: glocaltokens_client.get_google_devices_information(
-            zeroconf_instance
-        ),
+        update_method=lambda: glocaltokens_client.update_google_devices_information(),
```
KapJI, line 59:
```suggestion
        update_method=glocaltokens_client.update_google_devices_information,
```

custom_components/ha-google-home/api.py
```diff
@@ -110,16 +130,15 @@ async def get_alarms_and_timers(self, url, device):
...
+        """Retrieves devices from google home devices"""
 
         _LOGGER.debug("Fetching sensor data...")
 
         offline_devices = []
-        devices = await self.get_google_devices(zeroconf_instance)
         coordinator_data = {}
 
-        for device in devices:
+        for device in await self.get_google_devices():
```
KapJI, line 141: Processing awaitables in a loop is bad practice. Create async function to fetch alarms for single device and then use `asyncio.gather` to run it for all devices in parallel:
```python
devices = await self.get_google_devices(zeroconf_instance)
coordinator_data = await asyncio.gather([fetch_alarms(device) for device in devices]
```

custom_components/ha-google-home/const.py
```diff
@@ -69,7 +69,7 @@
 
 # Access token only lives about 1 hour
 # Update often to fetch timers in timely manner
-UPDATE_INTERVAL = 60 * 1  # every minute
+UPDATE_INTERVAL = 15  # sec
```
KapJI, line 72: How about 10 seconds? Local queries are fast and UI will update quicker.

KapJI (changes requested):

custom_components/ha-google-home/api.py
```diff
@@ -93,6 +105,14 @@ async def get_alarms_and_timers(self, url, device):
...
             if response.status == HTTP_OK:
                 resp = await response.json()
+            elif response.status == HTTP_UNAUTHORIZED:
+                # If token is invalid - force reload homegraph providing new token and rerun the task
+                _LOGGER.debug(
+                    "Failed to fetch data from {device} due to invalid token. Will refresh the token and try again".format(
+                        device=device.device_name,
+                    )
+                )
+                self.google_devices = []
```
KapJI, line 115: This means we won't have updated results for this update cycle and will only get it after `UPDATE_INTERVAL`. It's better to retry immediately.

custom_components/ha-google-home/api.py
leikoilja: uuu, that's a really nice improvement, i was wondering how to shoot those tasks in parallel, thanks a lot! :boom:

custom_components/ha-google-home/api.py
leikoilja: Yeh, but tbh i don't know how to retrigger the entire coordinator update method from here. I kind of assumed waiting 10 sec (UPDATE INTERVAL) once a day might not be too big of a problem. Thou, if can see any good way to terminate the current look and create a new async task to update coordinator - that would be very helpful

custom_components/ha-google-home/__init__.py
leikoilja: Wasn't able to drop `lambda`, it was throwing the error

leikoilja:
@KapJI, please take a look at the latest changes. :+1:

custom_components/ha-google-home/api.py
KapJI: Let's keep it for other PR then.

custom_components/ha-google-home/__init__.py
KapJI: What error? Make sure you're passing the function itself, not calling it.

leikoilja: Oh, flip you are totally right, i was calling the function. Thanks for the spot

custom_components/ha-google-home/api.py
```diff
@@ -128,55 +130,26 @@ async def get_alarms_and_timers(self, url, device):
...
-        _LOGGER.debug("Fetching sensor data...")
-
-        offline_devices = []
-        coordinator_data = {}
-
-        for device in await self.get_google_devices():
-            timers = []
-            alarms = []
-
-            if device.ip:
```
KapJI, line 126: It's still possible that device is offline.

KapJI (changes requested):

custom_components/ha-google-home/const.py
```diff
@@ -36,6 +36,18 @@
...
+SUPPORTED_HARDWARE_LIST = [
+    "Google Home Hub",
+    "Google Home Max",
+    "Google Home Mini",
+    "Google Nest Mini",
+    "Nest Audio",
+    "Nest Hub Max",
+    "Nest Hub",
+    "Nest Mini",
+]
```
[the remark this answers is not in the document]
leikoilja: Good spot, missed it

custom_components/ha-google-home/api.py
leikoilja: Wanted to move into a separate PR, but hey-ho, let's just keep it here :)

leikoilja:
Thanks for good comment, @KapJI. Pushed the changes, please take a look :rocket:

custom_components/ha-google-home/api.py
```diff
@@ -108,56 +132,40 @@ async def get_alarms_and_timers(self, url, device):
...
                     "Failed to fetch timers/alarms information from device {device}. We could not determine it's IP address, the device is either offline or is not compatable Google Home device. Will try again later.".format(
                         device=device.device_name
                     )
                 )
+
+        coordinator_data = await gather(
+            *[
+                self.get_alarms_and_timers(device)
+                for device in devices
+                if (device.ip and device.hardware in SUPPORTED_HARDWARE_LIST)
```
KapJI, line 167:
```python
if device.ip and device.hardware in SUPPORTED_HARDWARE_LIST
```

custom_components/ha-google-home/const.py
```diff
@@ -36,6 +36,19 @@
...
+SUPPORTED_HARDWARE_LIST = [
+    "Google Home Hub",
+    "Google Home Max",
+    "Google Home Mini",
+    "Google Home",
+    "Google Nest Mini",
+    "Nest Audio",
+    "Nest Hub Max",
+    "Nest Hub",
+    "Nest Mini",
```
KapJI, line 49: Sort it.
```python
    "Google Home",
    "Google Home Hub",
    "Google Home Max",
    "Google Home Mini",
    "Google Nest Mini",
    "Nest Audio",
    "Nest Hub",
    "Nest Hub Max",
    "Nest Mini",
```

custom_components/ha-google-home/manifest.json
```diff
@@ -2,7 +2,7 @@
   "after_dependencies": ["cloud", "http", "zeroconf"],
   "codeowners": ["@leikoilja", "@DurgNomis-drol", "@ArnyminerZ"],
   "config_flow": true,
-  "dependencies": [],
+  "dependencies": ["rest"],
```
KapJI, line 5: Where is it used?

KapJI (approved):

leikoilja:
Awesome, thanks @KapJI for reviews 💥

PR #64, opened 2021-03-07
Parse labels to timers and alarms

DurgNomis-drol:
Parses the `label` attribute to make it possible to show the name/label of the timer/alarm when it has one.

### DIFF
--- a/custom_components/ha-google-home/const.py
+++ b/custom_components/ha-google-home/const.py
@@ -53,6 +53,7 @@ TIMEOUT = 10  # Request Timeout in seconds
 
 # TIMERS & ALARMS ATTRIBUTE NAMES
 ID = "id"
+LABEL = "label"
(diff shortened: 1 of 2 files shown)


leikoilja (approved):
Wait what, you can set a label for a timer/alarm? :smile:  How do you do that?

DurgNomis-drol:
😆 

You can say `with the name 'Pizza' afterwards.` But there are many ways you can fraise the that so it picks it up. I also suspect it is very different for each language how many ways there is 😄

leikoilja:
haha, that's nice, i didn't know! Thanks for adding it, that would be so useful! :smile:  Was cooking yesterday and using two timers was confusing cause i didn't know which is which :smile:

DurgNomis-drol:
I know the feeling 😄

PR #65, opened 2021-03-07
Add next alarm sensor, and sort alarms and timers

DurgNomis-drol:
As described. This sorts alarms and timers and adds a new sensor called sensor.*_next_alarm which only shows the next alarm and it's attributes.

### DIFF
--- a/custom_components/ha-google-home/entity.py
+++ b/custom_components/ha-google-home/entity.py
@@ -118,7 +118,6 @@ class GoogleHomeNextAlarmEntity(CoordinatorEntity):
         return ICON_ALARMS
 
 
-
 class GoogleHomeTimersEntity(CoordinatorEntity):
     def __init__(self, coordinator, config_entry):
         super().__init__(coordinator)
--- a/custom_components/ha-google-home/entity.py
+++ b/custom_components/ha-google-home/entity.py
@@ -117,10 +117,6 @@ class GoogleHomeNextAlarmEntity(CoordinatorEntity):
         """Icon to use in the frontend, if any."""
         return ICON_ALARMS
 
-    @property
-    def state(self):
-        """Return the state of the sensor."""
-        return self._state
 
 
 class GoogleHomeTimersEntity(CoordinatorEntity):
--- a/custom_components/ha-google-home/entity.py
+++ b/custom_components/ha-google-home/entity.py
@@ -7,6 +7,7 @@ from .const import ICON_TIMERS
 from .const import ICON_TOKEN
 from .const import LABEL_ALARMS
 from .const import LABEL_NEXT_ALARM
+from .const import LABEL_NEXT_TIMER
 from .const import LABEL_TIMERS
 from .const import LABEL_TOKEN
 from .const import MANUFACTURER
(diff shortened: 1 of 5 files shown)


custom_components/ha-google-home/sensor.py
```diff
@@ -88,6 +94,51 @@ def _get_alarms_data(self):
...
+        super().__init__(coordinator, entry)
+        self._name = device_name
+
+    @property
+    def state(self):
+        alarms = self._get_alarm_data()
+        state = alarms[0][LOCAL_TIME_ISO] if len(alarms) else STATE_OFF
+        # The first one will always be the closet one
+        # as we have sorted the list in _get_alarm_data()
+        return state
```
KapJI, line 115: Multiple changes here:
```suggestion
        # The first one will always be the closest one
        # as we have sorted the list in _get_alarm_data()
        return alarms[0][LOCAL_TIME_ISO] if alarms else STATE_OFF
```

custom_components/ha-google-home/entity.py
```diff
@@ -87,6 +88,41 @@ def state(self):
...
+
+    @property
+    def icon(self):
+        """Icon to use in the frontend, if any."""
+        return ICON_ALARMS
+
+    @property
+    def state(self):
+        """Return the state of the sensor."""
+        return self._state
```
KapJI, line 123: This is already defined in `GoogleHomeNextAlarmSensor`
```suggestion
```

KapJI (changes requested):

custom_components/ha-google-home/entity.py
```diff
@@ -87,6 +88,37 @@ def state(self):
...
+        """Return a unique ID to use for this entity."""
+        return f"{self._name}/{LABEL_NEXT_ALARM}"
+
+    @property
+    def icon(self):
+        """Icon to use in the frontend, if any."""
+        return ICON_ALARMS
+
+
+
```
KapJI, line 121: Now there're too many blank lines 🙂

KapJI (approved):
Looks good to me!

leikoilja:
This is sooo cool, thanks guys for picking it up! 👍 
@DurgNomis-drol have you tried the sensor with multiple alarms per device where some are recurring ones? I'm wondering if `fire_time` is dynamically changed to which alarm is the soonest on google device that we can rely on?

Also, now that we are creating next_alarm, should we maybe consider making next_timer as well? We can reuse most of logic, but it would be consistent and probably useful as well?

DurgNomis-drol:
Thanks @leikoilja 😄 

I believe it does change fire_time dynamically, but it is something i we need to confirm before merging

It would be very simple to add a next timer sensor as well, I will add it as soon as you have rebased it 😄

leikoilja:
@DurgNomis-drol, you can try adding the next timer sensor now :) I'll test a little bit in the meanwhile.

DurgNomis-drol:
It does dynamically update `fire_time` in about 1 min or so.

leikoilja:
That's exactly what we need! 🚀  I have also just been testing and it seems to work really good! 💥

leikoilja:
@DurgNomis-drol do you mind updating the `README` to add there information about next alarm/timer as well? :)

DurgNomis-drol:
@leikoilja Good to here! 🚀 

And yes, will do it as the last thing before we merge this one 😄

leikoilja (changes requested):
Will request changes just to keep in loop :) Please re-request when we are ready to merge 💥

DurgNomis-drol:
And wow the integration got fast with the latest PR 😄 This is so awesome!

custom_components/ha-google-home/sensor.py
```diff
@@ -32,6 +36,12 @@ async def async_setup_entry(hass, entry, async_add_devices):
                             device.device_name,
                             getattr(device, LABEL_ALARMS),
                         ),
+                        GoogleHomeNextAlarmSensor(
+                            coordinator,
+                            entry,
+                            device.device_name,
+                            getattr(device, LABEL_ALARMS),
+                        ),
```
leikoilja, line 44: You are forgetting to initialize the next timer sensor here :)

leikoilja (approved):
Awesome stuff! Thanks for implementing it! 💥 

Please see my small comment. 
Also please **test that it works** and if so - let's merge it in 🚀 
It's a good habit to test something you have implementing before merging. Some of us that review can only spot some technicalities without testing if it works as excepted. And once it's merged it can potentially break `master` if not tested well(obv not in this particular implementation). And breaking `master` would not be good 😅 
Thanks again 💯

DurgNomis-drol:
I don't know why it disappeared, I added it, because i tested it earlier.

I always test, even before committing to my own branch. 😄  But i haven recently moved to my Windows 10 Desktop from a Linux laptop and Windows does not play well with pre-commit.

Sorry for any problems or confusion it have made. I will try to be better at spotting if something happens that should not happen in the future. 😄

leikoilja:
Oh, no worries at all! You are doing an awesome work 💥 keep it up 🚀

Issue #66, opened 2021-03-07
Use poetry for dependency management


KapJI:
[Poetry](https://python-poetry.org/) is a modern way to manage python projects. Right now dependency management is a bit messy with 2 separate dependency lists in `requirements` and `.github/workflows/constraints.txt`. And you need to manually create virtualenv if you don't want to mess with system packages.
- Poetry can install all project dependencies in virtualenv with `poetry install` and it manages dependency versions.
- It separates between normal dependencies and dev dependencies (black, flake8, etc)
- Then you can run any tools with `poetry run`, e.g. `poetry run black`, can be integrated with `pre-commit` as well.
- Some extra configuration is required for `dependabot`.

I want to switch everything (pre-commit config, github actions) to use poetry.
Any objections?

leikoilja:
No objections at all, not hand down 🤩💥
Never user poetry but I would love to hear and learn more about it! Thanks for taking an initiative with it 💥🤩

DurgNomis-drol:
Same opinion as @leikoilja :smile: 

And it looks very simple and easy use

PR #67, opened 2021-03-07
Remove unused configs

KapJI:
Preparation to #66 
- `isort` is not used, there is `reorder-python-imports` instead of it. Not sure which one is better.
- There are no tests yet, testing can enabled when we have them. No need in running empty github action before that.
- `tox` therefore is also not used and `poetry` can replace it for testing different configurations on github actions.


leikoilja (approved):
Good clean up, thank you

PR #68, opened 2021-03-07
Use poetry for dependency management

### LINKED ISSUE #66: Use poetry for dependency management

KapJI:
Remove old requirements files and create `pyproject.yaml` with poetry settings.
- Keep `0.0.0` as a placeholder, we're not going to build package for PyPI anyway. Component version should be placed in `manifest.json`.
- Change Python version to 3.8, as it's [current required version for home assistant](https://github.com/home-assistant/core/blob/79b5ca9415a9977f6f030d055149f57645e57d3e/homeassistant/const.py#L7). If you want, we can run lints for both 3.8 and 3.9.

### How to use poetry
- Install Poetry using pipx, pip or homebrew.
- Run `poetry install` to setup environment.
- (optional) Run `poetry run pre-commit install` to setup `pre-commit` git hook.
- Then you can run `poetry run pre-commit run --all-files` or just `git commit` (if you installed git hook).
- New dev dependencies can be added with `poetry add --dev flake8`, normal dependencies still need to be specified in `manifest.json`.

Closes #66


KapJI:
@leikoilja @DurgNomis-drol please take a look 🙂

poetry.lock
```diff
@@ -0,0 +1,438 @@
...
+colorama = ["colorama (>=0.4.3)"]
+d = ["aiohttp (>=3.3.2)", "aiohttp-cors"]
+
+[[package]]
+name = "cached-property"
+version = "1.5.2"
+description = "A decorator for caching properties in classes."
+category = "dev"
+optional = false
+python-versions = "*"
```
leikoilja, line 48: Do we need it? Or we have a plan to set start utilizing it?

poetry.lock
```diff
@@ -0,0 +1,438 @@
...
+
+[[package]]
+name = "nodeenv"
+version = "1.5.0"
+description = "Node.js virtual environment builder"
+category = "dev"
+optional = false
+python-versions = "*"
+
```
leikoilja, line 129: Also not very sure why we need this one 🤔

pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = "ha-google-home"
+version = "0.0.0"
```
leikoilja, line 3: Why do we set version to 0.0.0? Will it have to correspond to package version when we do a release?

leikoilja (approved):
It looks good to me, but a pretty new concept. All the requirements moved from 'requirements' file are automatically migrated to poetry.lock, right? 
There are few packages that I am not sure we need or are using, but I guess we'll just leave them for now 🤷‍♂️😅 
Thanks a lot for making the dev setup much better for us. Will you please also update README in contribution section to add a few lines how one should operate and use poetry now. (Pretty much all you have already mentioned in the issue)

poetry.lock
KapJI: This is autogenerated file and we shoulnd't edit it by hand. All these packages are installed as dependencies for those are in `pyproject.yaml`.

pyproject.toml
KapJI: Otherwise we will need to keep them in sync, it's better when you need to change it only in one place.

leikoilja: That's a good point. I'm wondering I go there is GitHub action or some utility to manage the versioning across many files based on gh release number 😅

KapJI: There is a plugin for poetry which can use git tags: https://github.com/mtkennerly/poetry-dynamic-versioning
But Home Assistant components are not regular python packages, so it's not very useful here. You can try it for `glocaltokens` though.

DurgNomis-drol:
I don't no anything about dependency management, but it looks very easy to use and maintain :smile: 

I just follow along reading what both of you are talking about and learning. It is approved from my perspective :+1: 

Thanks @KapJI for introducing us to it :rocket:

pyproject.toml
leikoilja: Oh perfect, that will be a good place to practice! Thanks man 🤩💥

PR #73, opened 2021-03-08
Use more Markdown in README

KapJI:
Let's keep html only for places where there is no Markdown alternative, e.g. centered header.
There should be any visual changes.

Note that empty line is required to use markdown again after html.


DurgNomis-drol (commented):
Awesome work! @KapJI

PR #76, opened 2021-03-08
Enable pylint

KapJI:
Pylint has many additional checks and it works well together with flake8.
I disabled all warnings which are failing right now, feel free to fix them and remove corresponding lines from `pyproject.toml`. Ideally that list should be empty.

This adds runtime dependencies: `zeroconf`, `homeassistant` and `glocaltokens`. They're needed for correct import resolution and later will be used for typing as well. Unfortunately because of old `grpcio-tools` version they take quite long to install (it builds from sources) so `poetry install` will be significantly slower. It doesn't affect CI job though, it seems it has wheels for ubuntu but not for mac.

Also update versions of pre-commit packages.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.3.0
+    rev: v3.4.0
     hooks:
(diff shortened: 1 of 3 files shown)


leikoilja (approved):
Perfect 🤩💥 thanks

PR #77, opened 2021-03-09
Removing SUPPORTED_HARDWARE_LIST

### LINKED ISSUE #71: Get rid of SUPPORTED_HARDWARE_LIST
# Problem
Maintaining that list is burden you don't really want to have. New devices will be released and other manufacturers will release compatible devices.
# Proposal
If device is discovered from Foyer API, it's probably good. Then we can check if it returns alarms list correctly, otherwise mark sensor `unavailable`.

leikoilja:
Removing supported_hardware_list and adding a few checks to ignore devices that are not available due to missing IP or not having timers/alarms capabilities.

Closes #71


custom_components/google_home/api.py
```diff
@@ -111,14 +111,22 @@ async def get_alarms_and_timers(self, device):
             elif response.status == HTTP_UNAUTHORIZED:
                 # If token is invalid - force reload homegraph providing new token and rerun the task
                 _LOGGER.debug(
-                    "Failed to fetch data from {device} due to invalid token. Will refresh the token and try again".format(
+                    (
+                        "Failed to fetch data from {device} due to invalid token. "
+                        "Will refresh the token and try again"
+                    ).format(
```
KapJI, line 117: Let's move long lines to #79

custom_components/google_home/api.py
```diff
@@ -111,14 +111,22 @@ async def get_alarms_and_timers(self, device):
...
                 self.google_devices = []
             elif response.status == HTTP_NOT_FOUND:
+                setattr(device, LABEL_AVAILABLE, False)
                 _LOGGER.debug(
-                    "Failed to fetch data from {device}, API returned {code}: The device(hardware='{hardware}') is not Google Home compatable and has no alarms/timers".format(
+                    (
+                        "Failed to fetch data from {device}, API returned {code}. "
+                        "The device(hardware='{hardware}') is not Google Home compatable and has no alarms/timers. "
+                        "We will not create sensor(s) or update {device} data anymore."
+                    ).format(
```
KapJI, line 129: Let's merge #79 and only add new part here.

custom_components/google_home/sensor.py
```diff
@@ -26,49 +26,41 @@ async def async_setup_entry(hass, entry, async_add_devices):
...
-                    ]
-                )
-            else:
-                _LOGGER.warning(
-                    "The {device} device(hardware='{hardware}') is not Google Home compatable and has no alarms/timers".format(
-                        device=device.device_name,
-                        hardware=device.hardware,
-                    )
-                )
+        if device.local_auth_token and getattr(device, LABEL_AVAILABLE, True):
```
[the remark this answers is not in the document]
leikoilja: Really good catch. I ll split this check into to and create token sensor for all, while alarm/timer sensors only for the AVAILABLE ones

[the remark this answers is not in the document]
KapJI: Please still add them in a single `async_add_devices` call.

custom_components/google_home/api.py
```diff
@@ -116,18 +142,20 @@ async def get_alarms_and_timers(self, device):
                         "Failed to fetch data from {device} due to invalid token. "
                         "Will refresh the token and try again."
                     ).format(
-                        device=device.device_name,
+                        device=device.name,
                     )
                 )
+                # TODO: Retry the update task
```
KapJI, line 148: Pylint doesn't like TODO comments, you can add "fixme" to the list of disabled checks.

KapJI:
@leikoilja why don't you use `pre-commit` locally? It can significantly speed up your workflow.

leikoilja:
I do, i do, @KapJI, give me 30-40 mins, it's still WIP. Just needed to move my changes from mac to linux machine for testing :D i ll re-request changes when ready

KapJI:
Take your time, there is no hurry 🙂

DurgNomis-drol:
I will just be lurking 👀 and learning, just say if there's is anything I can help with 😄

leikoilja:
Nah, started refactoring some other places and shot myself in the leg, can't find what i messed up 😄 Will continue tomorrow and ping you guys :)

KapJI:
I'd suggest to not go too deep into refactorings in one PR and split it in smaller chunks. This way it'll be easier to review and merge.

custom_components/google_home/api.py
```diff
@@ -19,14 +20,31 @@
...
 
 
+class GoogleHomeDevice:
+    def __init__(
+        self,
+        name: str,
+        token: str,
+        ip: Optional[str] = None,
+        hardware: Optional[str] = None,
+    ) -> None:
```
KapJI, line 37:
```suggestion
    ):
```

custom_components/google_home/api.py
```diff
@@ -19,14 +20,31 @@
...
+        """Local representation of GoogleHome device"""
```
KapJI, line 38:
```suggestion
        """Local representation of Google Home device"""
```

custom_components/google_home/sensor.py
```diff
@@ -95,14 +86,14 @@ def get_device(self):
 class GoogleHomeTokenSensor(GoogleHomeSensorMixin, GoogleHomeTokenEntity):
     """GoogleHome Sensor class."""
 
-    def __init__(self, coordinator, entry, device_name, local_auth_token):
+    def __init__(self, coordinator, entry, name, token):
```
KapJI, line 89: Let's keep `device_name` and `auth_token` for arguments. `name` is too generic IMO in this context.

custom_components/google_home/api.py
```diff
@@ -19,14 +20,31 @@
...
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
 
+class GoogleHomeDevice:
+    def __init__(
+        self,
+        name: str,
+        token: str,
+        ip: Optional[str] = None,
```
KapJI, line 35:
```suggestion
        ip_address: Optional[str] = None,
```

custom_components/google_home/api.py
```diff
@@ -19,14 +20,31 @@
...
 from .exceptions import InvalidMasterToken
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
 
+class GoogleHomeDevice:
+    def __init__(
+        self,
+        name: str,
+        token: str,
```
KapJI, line 34:
```suggestion
        auth_token: str,
```

custom_components/google_home/sensor.py
leikoilja: But is it? :thinking: 
When it comes to a `device` object of this specific class why to repeat ourselves saying `device_name` and refer to it as `device.device_name` were simply `device.name` is quite elaborate

leikoilja:
@KapJI, yeeeh, that what an experienced person would do - split into smaller chunks and PRs, but i got too carried away and refactored the :hankey: out of it :smile: 
Sorry guys. I'll do my best to teach myself to split into smaller chunks further down. Hope i didn't miss anything important

custom_components/google_home/sensor.py
KapJI: `device.name` is good. But here you don't have `device` anymore, I'm talking only about sensor here.
And I think `auth_token` is a little nicer than just `token` because there're different kinds of tokens which may exist.

custom_components/google_home/models.py
```diff
@@ -0,0 +1,137 @@
...
+            _dict.pop("_alarms")
+        return _dict
+
+
+class GoogleHomeTimer:
+    """Local representation of Google Home timer"""
+
+    def __init__(
+        self,
+        _id,
```
KapJI, line 109: Shouldn't start with `_`.
```suggestion
        timer_id,
```

custom_components/google_home/models.py
```diff
@@ -0,0 +1,137 @@
...
+        fire_time,
+        duration,
+        label,
+    ) -> None:
+        self._id = _id
```
KapJI, line 114: Don't you want to give access to it? `self._timer_id` should mean that this member is protected. If you just don't want to override built-in `id` function, please use `timer_id`.
```suggestion
        self._timer_id = timer_id
```

custom_components/google_home/models.py
```diff
@@ -0,0 +1,137 @@
...
+        dt_local = as_local(dt_utc)
+        self.local_time = dt_local.strftime(DATETIME_STR_FORMAT)
+        self.local_time_iso = dt_local.isoformat()
+
+
+class GoogleHomeAlarm:
+    """Local representation of Google Home alarm"""
+
+    def __init__(self, _id, fire_time, label, recurrence) -> None:
+        self._id = _id
```
KapJI, line 129: Same here. `alarm_id`

custom_components/google_home/models.py
```diff
@@ -0,0 +1,137 @@
...
+            GoogleHomeTimer(
+                _id=timer[ID],
+                fire_time=timer[FIRE_TIME],
+                duration=timer[ORIGINAL_DURATION],
+                label=timer.get(LABEL),
+            )
+            for timer in timers
+        ]
+
+    def get_alarms(self):
```
KapJI, line 65: How about this? It will be cleaner in places there this is used.
```suggestion
    def get_sorted_alarms(self):
```

custom_components/google_home/models.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @staticmethod
+    def as_dict(obj, flat=False):
+        """Returns object representation as dictionary
+        flat=True removes nested objects like list of alarms/timers"""
+        _dict = json.loads(json.dumps(obj, default=lambda o: o.__dict__))
+        if flat:
+            _dict.pop("_timers")
+            _dict.pop("_alarms")
+        return _dict
```
KapJI, line 101: This is **very** dangerous. Please remove this magic with dicts completely.
1. If `obj` is a list (alarms and timers), `json.dumps` will generate a list and this is an error.
2. There is no guarantee `dict.pop` will succeed, it may not be even a dict.
3. It's untyped

custom_components/google_home/sensor.py
```diff
@@ -29,96 +18,86 @@
...
     def get_device(self):
         """Return the device matched by name
         from the list of google devices in coordinator_data"""
         return next(
-            (
-                device
-                for device in self.coordinator.data
-                if device.device_name == self._name
-            ),
+            (device for device in self.coordinator.data if device.name == self._name),
```
KapJI, line 62: This should be a list.
```suggestion
            [device for device in self.coordinator.data if device.name == self._name],
```

KapJI (changes requested):

custom_components/google_home/models.py
leikoilja: haha, yeh, i was a bit unsure about this "magic" :D Can you help me to find a better way to serialize that nested object to the dictionary?

custom_components/google_home/sensor.py
leikoilja: Can iterator not be a tuple? 🤔

custom_components/google_home/sensor.py
leikoilja: Yeh, that's reasonable. Thanks, will change :)

custom_components/google_home/sensor.py
leikoilja: Nah, getting an error here if i change it to `list`:
```bash
    return next(
TypeError: 'list' object is not an iterator
```

custom_components/google_home/models.py
KapJI: Just construct `device_state_attributes` explicitly. If you really want to avoid writing the same parameters again, take a look at [TypedDict](https://www.python.org/dev/peps/pep-0589/)

custom_components/google_home/sensor.py
KapJI: Yeah, actually tuple generator is the right way.

custom_components/google_home/sensor.py
```diff
@@ -67,6 +67,23 @@ def get_device(self):
...
+            "hardware": device.hardware,
+            "available": device.available,
+            "integration": DOMAIN,
+        }
+
+    @staticmethod
+    def as_dict(obj_list):
+        """Return list of objects represented as dictionaries """
+        return [obj.__dict__ for obj in obj_list]
+
```
KapJI: Can we remove this as well?

custom_components/google_home/sensor.py
```diff
@@ -179,7 +196,7 @@ def device_state_attributes(self):
     def _get_timers_data(self):
         """Update timers data extracting it from coordinator"""
         device = self.get_device()
-        return device.get_sorted_timers_as_dict()
+        return self.as_dict(device.get_sorted_timers())
```
KapJI: It seems this can be removed. `"timers": device.get_sorted_timers()` should be enough.

custom_components/google_home/sensor.py
```diff
@@ -67,6 +67,23 @@ def get_device(self):
             None,
         )
 
+    @staticmethod
+    def get_device_attributes(device):
```
KapJI, line 71: Why not placing it directly in `GoogleHomeDeviceSensor`?

KapJI (changes requested):

custom_components/google_home/sensor.py
```diff
@@ -29,96 +18,107 @@
...
 
-    def __init__(self, coordinator, entry, device_name, local_auth_token):
+    def __init__(self, coordinator, entry, device_name, auth_token):
         """Initialize the sensor."""
         super().__init__(coordinator, entry)
-        self._name = device_name
+        self.device_name = device_name
+        self.auth_token = auth_token
+        device = self.get_device()
+        self.initial_attributes = self.get_device_attributes(device)
```
KapJI, line 85: Why do you need this? Can we return something like this if device is not available?
```
{
    "device_name": self.device_name,
    "auth_token": None,
    "ip_address": None,
    "hardware": None,
    "available": False,
    "integration": DOMAIN,
}
 ```

custom_components/google_home/sensor.py
KapJI: Yeah, I see. Let's keep it for now then.

custom_components/google_home/sensor.py
leikoilja: haha, valid point! sorry i m getting a bit exhausted here :D

KapJI (approved):
LGTM! Let's merge it! 🚀

PR #80, opened 2021-03-09
Update CONTRIBUTING.md

KapJI:
Add note regarding testing on MacOS.


leikoilja (approved):
Haven't seem that problem on my setup, but i guess that must be useful for others. Are you on amd64?

PR #83, opened 2021-03-10
Use lazy % formatting in logging functions

KapJI:
There is some controversy in python community about this check:
- https://github.com/PyCQA/pylint/issues/1788
- https://github.com/PyCQA/pylint/issues/2395

As I understand it, the main idea here is do not evaluate logging strings if they're not going to be printed.
I checked that Home Assistant follows this, so let's do the same.

@leikoilja sorry, you'll need to rebase one more time


leikoilja (approved):
Looks great, thanks :boom:

PR #88, opened 2021-03-11
Enable mypy

KapJI:
Add it to `pre-commit` config.

Unfortunately, `homeassistant`, `voluptuous` and `glocaltokens` don't distribute typing information.

I'll try to enable it for `homeassistant`.
@leikoilja can you enable this for `glocaltokens`? Should be one line in `setup.py`.
https://www.python.org/dev/peps/pep-0561/

### DIFF
--- a/custom_components/google_home/models.py
+++ b/custom_components/google_home/models.py
@@ -1,7 +1,9 @@
 """Models for Google Home"""
 
+from __future__ import annotations
+
 from datetime import timedelta
-from typing import List, Optional
+from typing import Any, Dict, List, Optional
 
 from homeassistant.util.dt import as_local, utc_from_timestamp
 
@@ -15,7 +17,7 @@ from .const import (
 )
 
 
-def convert_from_ms_to_s(timestamp):
+def convert_from_ms_to_s(timestamp: int) -> int:
     """Converts from milliseconds to seconds"""
     return round(timestamp / 1000)
 
@@ -29,7 +31,7 @@ class GoogleHomeDevice:
         auth_token: str,
         ip_address: Optional[str] = None,
         hardware: Optional[str] = None,
-    ) -> None:
+    ):
         self.name = name
         self.auth_token = auth_token
         self.ip_address = ip_address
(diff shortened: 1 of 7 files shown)


custom_components/google_home/models.py
```diff
@@ -1,7 +1,9 @@
 """Models for Google Home"""
 
+from __future__ import annotations
```
KapJI, line 3: This is for using forward declared types in type hints: https://www.python.org/dev/peps/pep-0563/

KapJI:
Feel free to add type hints to more parts after you merge this 🙂

KapJI:
@leikoilja can you take a look? 🙂

custom_components/google_home/models.py
leikoilja: hmmmm, dont really follow what it's going to do 🤔

setup.cfg
```diff
@@ -14,3 +14,15 @@ ignore =
...
+python_version = 3.8
+
+[mypy-glocaltokens.*]
+ignore_missing_imports = True
+
+[mypy-homeassistant.*]
+ignore_missing_imports = True
+
+[mypy-voluptuous.*]
+ignore_missing_imports = True
```
leikoilja, line 28: Why is it needed?

leikoilja (approved):
In general looks really good to me! Thanks for making it! I have a few conceptual questions, but those are mostly around my limited understanding of a topic :) If you could help clearing those up, i'd be happy

custom_components/google_home/models.py
KapJI: Basically this allows you to use types before they're defined.
For example here it is:

```python3
def get_next_alarm(self) -> Optional[GoogleHomeAlarm]:
```

Otherwise you need to use string:

```python3
def get_next_alarm(self) -> Optional["GoogleHomeAlarm"]:
```

Should be enabled by default in Python 3.10.

setup.cfg
KapJI: These packages don't distrubute typing information (don't have `py.typed` marker) and `mypy` will complain that it can't import these packages.

leikoilja: Got it, thank you!
Check `glocaltokens` open PR, i think it's ready to go. I have added `py.typed` as requested :)

custom_components/google_home/models.py
leikoilja: uuu that't nice, thanks, i hope i get it a bit better now

setup.cfg
KapJI: Yes, with it we can remove `glocaltokens` from here and type check its usages :)

PR #92, opened 2021-03-13
Add more type hints

KapJI:
- Callback `Protocol` really should be a part of home assistant, but they don't have it. Maybe will add it to the stubs package in the future.
- `Any` will be banned at some point, but this requires reorganising the code, keep it for another PR.


leikoilja (approved):
Awesome stuff, thanks for adding type hints! 💯

Issue #102, opened 2021-03-22
No sensors created - SSL handshake failure


Duncank:
This might be similar to #100, but since the error message differs I've created a new issue for this.

## Version of the custom_component
1.0.0

## Configuration
via `integrations` panel in UI.

## Describe the bug
No sensors are created after setting up the integration and rebooting a few times.

## Debug log
```
Logger: custom_components.google_home
Source: helpers/update_coordinator.py:160
Integration: Google Home (documentation, issues)
First occurred: 12:53:12 (8 occurrences)
Last logged: 13:01:03

Error requesting sensor data: Cannot connect to host 192.168.2.201:8443 ssl:default [[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1125)]
```

The issue stays after reboot, sensors are not created. I've installed the latest version of this integration via HACS on a Raspberry Pi 3b.

DurgNomis-drol:
Can you press the button `Full Home Assistant logs` and see if there is a `traceback` that could gives us some more info? The message is not generated directly from our `integration` but by some of the `libs` we use. (In this case probably `aiohttp`).

Duncank:
I'm not seeing lines in the log that give me more information. Most of the lines are for my zigbee integration, and within 60 seconds before and after the error I mentioned above, there is nothing that is not related to zigbee.

This is the full line for one instance of the errors.
`2021-03-22 15:24:00 ERROR (MainThread) [custom_components.google_home] Error requesting sensor data: Cannot connect to host 192.168.2.201:8443 ssl:default [[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1125)]
`

`aiohttp` is not found in the log at all.

DurgNomis-drol:
You wont find `aiohttp` in the logs as it is an internal component.

Can you please add this to your `configuration.yaml`:

```yaml
logger:
  default: info
  logs:
     custom_components.google_home: debug
     glocaltokens: debug
```

This will enable us to get a better picture of why the integration fails. 😄 

Thanks for your feedback.

Duncank:
Here we go! I see 5 devices for which the integration is trying to fetch data, but these are not all Google Home devices, so probably that is the problem here, just like in #100.  
`Zolder Mini`, `Keuken Speaker` and `Badkamer` are Google Home / Nest Home / Nest Audio devices.  
[...]

PR #103, opened 2021-03-22
Catch exceptions when getting alarms and timers to better allow for debugging

### LINKED ISSUE #100: Error requesting sensor data

1.0
## Configuration

```yaml
Add your logs here.
```

## Describe the bug

I've installed the custom component and I was able to add it in the integration page. However, I'm not getting the sensors in my HA. I might be missing something, but I don't know what. 
## Debug log

```text

Logger: custom_components.google_home
Source: helpers/update_coordinator.py:160
Integration: Google Home (documentation, issues)
First occurred: 10:53:42 (3 occurrences)
Last logged: 10:55:19

Error requesting sensor data: Cannot connect to host 192.168.xx.xx:8443 ssl:default [None]

```

DurgNomis-drol:
This should handle these to problems and allow for us in the future to better debug if the underlaying package is throwing an exceptions when retrieving data from each device. 

We need some one with problem to test it first so we can confirm that it solves the problem. I have tested it on my machine and it works.

Should close #100 and #102 

More info:

The error both of them gets, is an exception thrown by, in this case, `aiohttp` because those devices does not conform. We should just skip those devices and log them.

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -151,13 +151,13 @@ class GlocaltokensApiClient:
                         response.status,
                         response,
                     )
-        except Exception as exc:  # pylint: disable=broad-except
+        except Exception as ex:  # pylint: disable=broad-except
             # Make sure that we log the exception if one occurred.
             # The only reason we do this broad is so we easily can
             # debug the application.
             _LOGGER.error(
                 "Request error: %s",
-                exc,
+                ex,
             )
         finally:
             if resp:
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -151,7 +151,7 @@ class GlocaltokensApiClient:
                         response.status,
                         response,
                     )
-        except Exception as ex:  # pylint: disable=broad-except
+        except aiohttp.ClientError as ex:
             # Make sure that we log the exception if one occurred.
             # The only reason we do this broad is so we easily can
             # debug the application.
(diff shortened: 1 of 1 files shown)


DurgNomis-drol:
@leikoilja @KapJI What do you guys think?

custom_components/google_home/api.py
```diff
@@ -113,51 +113,63 @@ async def get_alarms_and_timers(self, device):
...
+                        device.hardware,
+                    )
+                else:
+                    _LOGGER.error(
+                        "Failed to fetch %s data, API returned %d: %s",
+                        device.name,
+                        response.status,
+                        response,
+                    )
+        except Exception as exc:  # pylint: disable=broad-except
```
KapJI, line 154: `ex` is more common here. If pylint complains about it, in `pyproject.toml` add:
```
[tool.pylint.basic]
good-names = ["ex"]
```

KapJI: Also catching all possible exceptions in this large block of code is really too broad. I think you only need `aiohttp.ClientError` here.

KapJI (changes requested):

custom_components/google_home/api.py
DurgNomis-drol: I know it is bad to catch broad, but i specifically thought that in this case it will be pretty useful as a debug tool for us.

But it is possible to use another client library then `aiohttp`?

DurgNomis-drol: No you are right, it is not compatible, I forgot that `aiohttp` is a little bit different then others in its api.

DurgNomis-drol:
[Confirmed](https://github.com/leikoilja/ha-google-home/issues/100#issuecomment-804244668) to solve the issue so far.

KapJI (approved):
Looks good, thank you!

leikoilja:
Yay 🥳 thank you guys for taking care of it 💥🤩

PR #107, opened 2021-03-23
Add device class for next alarm and next timer.

### LINKED ISSUE #105: Next alarm/timer should be device_class timestamp
I would like to use some blueprints that use timestamps but was surprised to find that I could not use the next alarm or or timers  sensor because they are not device_class timestamp. Since these sensor provide a timestamp, they should have this device class

DurgNomis-drol:
Should close #105 

This was my fault when I first implemented next alarm and next timer sensors.

### DIFF
--- a/custom_components/google_home/entity.py
+++ b/custom_components/google_home/entity.py
@@ -66,7 +66,7 @@ class GoogleHomeDeviceEntity(CoordinatorEntity):
 
     @property
     def device_class(self) -> str:
-        """Return de device class of the sensor."""
+        """Return the device class of the sensor."""
         return "google_home__custom_device_class"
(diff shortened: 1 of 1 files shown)


leikoilja (approved):
Haha, don't blame yourself, not a biggie at all! Thanks for implementing the fix! 💥 
Just out of curiosity, have you been able to test it on your HA setup to make sure it works? :) Would be nice to make sure it actually returns the expected result before we merge it in :)

DurgNomis-drol:
@leikoilja I just tested it and it works 😄

leikoilja (approved):
Legend! Thanks man! 💥

PR #110, opened 2021-03-25
Set verbose logging level for glocaltokens

leikoilja:
Setting verbose logging level for glocaltokens based on integration logging level.

Also prepare for release by bumping up the version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,10 +24,12 @@ mypy = "^0.812"
 # Reasons disabled:
 # too-many-* - not enforced for the sake of readability
 # too-few-* - same as too-many-*
+# unexpected-keyword-arg - argues that glocaltokens has no verbose argument
 disable = [
     "too-few-public-methods",
     "too-many-arguments",
     "too-many-instance-attributes",
+    "unexpected-keyword-arg"
 ]
 
 [tool.pylint.format]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -24,10 +24,12 @@ mypy = "^0.812"
 # Reasons disabled:
 # too-many-* - not enforced for the sake of readability
 # too-few-* - same as too-many-*
+# unexpected-keyword-arg - argues that glocaltokens has no verbose argument
```
KapJI, line 27: It's a valid complain. You need to update `glocaltokens` version in poetry as well.

leikoilja: yeh, it felt like i m missing something here :D

leikoilja: Fixed ✅

KapJI:
Hm. I think version constraint in `pyproject.toml` also need to be changed. Right now it's `0.2.4 or above` which is not right. Try `poetry add glocaltokens==0.2.6`.

leikoilja:
Hm, that's weird. I have ran `poetry update glocaltokens` I would have expected it would update the version in `pyproject.toml` file as well. Needed to run `remove` and `add glocaltokens` to reinstall it to the newest version 🤔

KapJI:
Yeah, `poetry update` only updates version in `poetry.lock` using current constraints from `pyproject.toml`.
So right now it's `0.2.6 or above` which is good but let's maybe keep constraint in sync with `manifest.json` and set fixed version here as well?

leikoilja:
Hmmmm, sounds good with me :)
Fixed

KapJI (approved):
Thanks! 🎉

PR #111, opened 2021-03-25
Reduce code duplication in entities

KapJI:
Simplify class structure.
- One base class for common code, remove mixin.
- All specialisation is done on sensor level.
- Remove `model` with invalid `VERSION` for `device_info`. We don't really want to maintain version in different places.
- Make `GoogleHomeDeviceSensor` constructor the same as for other sensors. If there is no such device in `coordinator_data`, set state to `None`, it should render as `unknown` in HA.
- Rename `GoogleHomeAlarmSensor` to `GoogleHomeAlarmsSensor`, same for `Timers`.
- Delete non existing device class for `GoogleHomeDeviceSensor`.
- Update docstrings.

### DIFF
--- a/custom_components/google_home/entity.py
+++ b/custom_components/google_home/entity.py
@@ -1,27 +1,13 @@
 """Defines base entities for Google Home"""
 
-from typing import Set, Tuple, TypedDict
+from typing import Any, Dict, List, Set, Tuple, TypedDict
 
-from homeassistant.const import DEVICE_CLASS_TIMESTAMP
 from homeassistant.helpers.update_coordinator import (
     CoordinatorEntity,
     DataUpdateCoordinator,
 )
 
-from .const import (
-    DEFAULT_NAME,
-    DOMAIN,
-    ICON_ALARMS,
-    ICON_TIMERS,
-    ICON_TOKEN,
-    LABEL_ALARMS,
-    LABEL_DEVICE,
-    LABEL_NEXT_ALARM,
-    LABEL_NEXT_TIMER,
-    LABEL_TIMERS,
-    MANUFACTURER,
-    VERSION,
-)
+from .const import DEFAULT_NAME, DOMAIN, MANUFACTURER
 
 
 class DeviceInfo(TypedDict):
@@ -29,12 +15,11 @@ class DeviceInfo(TypedDict):
 
     identifiers: Set[Tuple[str, str]]
     name: str
-    model: str
     manufacturer: str
 
 
-class GoogleHomeDeviceEntity(CoordinatorEntity):
-    """Entity base for device sensor"""
+class GoogleHomeBaseEntity(CoordinatorEntity):
+    """Base entity base for Google Home sensors"""
 
     def __init__(self, coordinator: DataUpdateCoordinator, device_name: str):
         super().__init__(coordinator)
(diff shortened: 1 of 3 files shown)


custom_components/google_home/entity.py
```diff
@@ -183,26 +30,22 @@ def device_info(self) -> DeviceInfo:
...
+        """Return the device matched by device name
+        from the list of google devices in coordinator_data"""
+        return next(
+            (
+                device
+                for device in self.coordinator.data
+                if device.name == self.device_name
+            ),
+            None,
+        )
```
leikoilja, line 46: I still think this iterator might be problematic and we would be safer with something simpler like `for .. in ...:` and returning when the device is found. But let's keep it for further investigations and PRs

custom_components/google_home/sensor.py
```diff
@@ -68,42 +70,28 @@ async def async_setup_entry(
...
-    def __init__(
-        self, coordinator: DataUpdateCoordinator, device_name: str, auth_token: str
-    ):
-        """Initialize the sensor."""
-        super().__init__(coordinator, device_name)
-        self.auth_token = auth_token
+    @property
+    def icon(self) -> str:
+        """Return the icon of the sensor."""
+        return ICON_TOKEN
```
leikoilja, line 89: How about we set on a class level:
```python
class GoogleHomeDeviceSensor(GoogleHomeBaseEntity):
  LABEL = LABEL_DEVICE
  ICON = ICON_TOKEN
  ....
```
and then move all those reperative `def name; def unique_id; def icon` back to `GoogleHomeBaseEntity` where these methods would look like:
```python
    @property
    def name(self) -> str:
        """Return the name of the sensor."""
        return f"{self.device_name} {self.LABEL}"
....
```
It would help us avoid repetitions and keep cleaner :)

custom_components/google_home/sensor.py
```diff
@@ -133,8 +121,23 @@ def get_device_attributes(device):
...
+
+    @property
+    def unique_id(self) -> str:
+        """Return a unique ID to use for this entity."""
+        return f"{self.device_name}/{LABEL_ALARMS}"
+
+    @property
+    def icon(self) -> str:
+        """Icon to use in the frontend."""
+        return ICON_ALARMS
```
leikoilja, line 140: Same as above

custom_components/google_home/sensor.py
```diff
@@ -156,8 +159,28 @@ def _get_alarms_data(self) -> List[Dict[Any, Any]]:
...
+
+    @property
+    def unique_id(self) -> str:
+        """Return a unique ID to use for this entity."""
+        return f"{self.device_name}/{LABEL_NEXT_ALARM}"
+
+    @property
+    def icon(self) -> str:
+        """Icon to use in the frontend."""
+        return ICON_ALARMS
```
leikoilja, line 178: Same as above

custom_components/google_home/sensor.py
```diff
@@ -183,8 +206,23 @@ def _get_next_alarm(self):
...
+
+    @property
+    def unique_id(self) -> str:
+        """Return a unique ID to use for this entity."""
+        return f"{self.device_name}/{LABEL_TIMERS}"
+
+    @property
+    def icon(self) -> str:
+        """Icon to use in the frontend."""
+        return ICON_TIMERS
```
leikoilja, line 225: Same as above

custom_components/google_home/sensor.py
```diff
@@ -205,8 +243,28 @@ def _get_timers_data(self) -> List[Dict[Any, Any]]:
...
+
+    @property
+    def unique_id(self) -> str:
+        """Return a unique ID to use for this entity."""
+        return f"{self.device_name}/{LABEL_NEXT_TIMER}"
+
+    @property
+    def icon(self) -> str:
+        """Icon to use in the frontend."""
+        return ICON_TIMERS
```
leikoilja, line 262: Same as above

leikoilja (changes requested):
Looks really good, thank you 💥 

The only thing I'm not really happy about is duplicating all the methods in sensors/alarms. But i think we can reduce that too :)

custom_components/google_home/entity.py
KapJI: Used normal list here.

custom_components/google_home/sensor.py
KapJI: Parent class can't really use constants defined in subclasses. But we can use abstract property here.

DurgNomis-drol:
Thanks @KapJI for taking the initiative to do this. 🚀😁

leikoilja:
@KapJI, are you still working on it? Please re-request review when you want some of us to review :)

custom_components/google_home/entity.py
```diff
@@ -1,208 +1,66 @@
...
-        return DEVICE_CLASS_TIMESTAMP
+    def get_device(self):
+        """Return the device matched by device name
+        from the list of google devices in coordinator_data"""
+        matched_devices = [
+            device
+            for device in self.coordinator.data
+            if device.name == self.device_name
+        ]
+        return matched_devices[0] if matched_devices else None
```
leikoilja, line 61: Love it! Thanks for implementing it. Maybe not the most looking solution, but hopefully more robust

leikoilja (approved):
Looks good to me! 
I'd be happy if we can try it locally just to make sure it doesn't break anything before merging it :)

KapJI:
I tested it locally, works as before :)

PR #112, opened 2021-03-25
Combine two sensors into one

### LINKED ISSUE #97: AttributeError: 'NoneType' object has no attribute 'get_sorted_alarms

Latest

## Configuration

```yaml
2021-03-21 09:41:29 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 126, in _handle_refresh_interval
    await self.async_refresh()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 200, in async_refresh
    update_callback()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 260, in _handle_coordinator_update
    self.async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 295, in async_write_ha_state
    self._async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 319, in _async_write_ha_state
    sstate = self.state
  File "/config/custom_components/google_home/sensor.py", line 141, in state
    alarms = self._get_alarms_data()
  File "/config/custom_components/google_home/sensor.py", line 156, in _get_alarms_data
    return self.as_dict(device.get_sorted_alarms())
AttributeError: 'NoneType' object has no attribute 'get_sorted_alarms'
```

## Describe the bug

It does not appear immediately, but shows up after my HA installation have been running for some hours. 

@leikoilja Have also had this error show up in his logs

### LINKED ISSUE #106: Alarm/timer sensors should be binary_sensor
The alarm/timer sensor provide a binary value (are there any active alarms/timer). It is odd that this information is not provided through a binary_sensor instead

KapJI:
State of alarms is now the next alarm. Same for timers. Set device class to `timestamp`.

Also if `device` is `None`, set state and attributes to some value instead of throwing an error (Fixes #97).

Update documentation and move alarms on the top, I think they're used more often.

Tested locally.

Closes #106

### DIFF
--- a/custom_components/google_home/sensor.py
+++ b/custom_components/google_home/sensor.py
@@ -1,11 +1,11 @@
 """Sensor platforms for Google Home"""
 import logging
-from typing import Any, Dict, Iterable, List
+from typing import Any, Dict, Iterable, List, Optional
 
 from typing_extensions import Protocol
 
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import DEVICE_CLASS_TIMESTAMP, STATE_OFF, STATE_ON
+from homeassistant.const import DEVICE_CLASS_TIMESTAMP, STATE_OFF
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity import Entity
 
@@ -16,8 +16,6 @@ from .const import (
     ICON_TOKEN,
     LABEL_ALARMS,
     LABEL_DEVICE,
-    LABEL_NEXT_ALARM,
-    LABEL_NEXT_TIMER,
     LABEL_TIMERS,
 )
 from .entity import GoogleHomeBaseEntity
@@ -53,15 +51,7 @@ async def async_setup_entry(
                     coordinator,
                     device.name,
                 ),
-                GoogleHomeNextAlarmSensor(
-                    coordinator,
-                    device.name,
-                ),
                 GoogleHomeTimersSensor(
-                    coordinator,
-                    device.name,
-                ),
-                GoogleHomeNextTimerSensor(
                     coordinator,
                     device.name,
                 ),
@@ -84,7 +74,7 @@ class GoogleHomeDeviceSensor(GoogleHomeBaseEntity):
         return ICON_TOKEN
 
     @property
-    def state(self) -> str:
+    def state(self) -> Optional[str]:
         device = self.get_device()
         return device.auth_token if device else None
 
@@ -100,8 +90,7 @@ class GoogleHomeDeviceSensor(GoogleHomeBaseEntity):
             "available": False,
             "integration": DOMAIN,
         }
-        attributes = self.get_device_attributes(device) if device else attributes
-        return attributes
+        return self.get_device_attributes(device) if device else attributes
 
     @staticmethod
     def get_device_attributes(device):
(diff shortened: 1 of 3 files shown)


custom_components/google_home/sensor.py
```diff
@@ -130,10 +119,17 @@ def icon(self) -> str:
...
-        return state
+    def device_class(self) -> str:
+        """Return the device class of the sensor."""
+        return DEVICE_CLASS_TIMESTAMP
+
+    @property
+    def state(self) -> Optional[str]:
+        device = self.get_device()
+        if not device:
+            return None
```
leikoilja, line 130: Good way of avoiding the error if device is not found 🤔 👍

leikoilja (approved):
I like that! Less code and cleaner. Hope it's gonna continue being as convenient as used to be `next_alarm` 😄 You won that one haha 😄 
Thanks for "Tested locally" 💥 

I'll add a label `minor`, since it's kind of a breaking change for some, so this will make it `v1.1.0` instead of default patch `v.1.0.2`

Let's also allow @DurgNomis-drol to check it and then merge merge merge! 🚀

custom_components/google_home/sensor.py
```diff
@@ -146,68 +142,34 @@ def device_state_attributes(self):
     def _get_alarms_data(self) -> List[Dict[Any, Any]]:
         """Update alarms data extracting it from coordinator"""
         device = self.get_device()
-        return self.as_dict(device.get_sorted_alarms())
+        return self.as_dict(device.get_sorted_alarms()) if device else []
```
DurgNomis-drol, line 145: We return `as_dict` but if not a list an empty?

custom_components/google_home/sensor.py
```diff
@@ -220,46 +182,4 @@ def device_state_attributes(self):
...
-            }
-        )
-        return attributes
-
-    def _get_next_timer(self):
-        """Update alarms data extracting it from coordinator"""
-        device = self.get_device()
-        timer = device.get_next_timer()
-        return timer
+        return self.as_dict(device.get_sorted_timers()) if device else []
```
DurgNomis-drol, line 185: We do the same here?

DurgNomis-drol (approved):
Looks good to me! See my questions, maybe it is just me that don't understand something about python. 😄

custom_components/google_home/sensor.py
leikoilja: Oh, maybe we should have better named it `as_list_of_dict` 😄 
image
We basically feed it either a list or empty list in this line and it returns a list of dictionaries or an empty list

KapJI: Yeah, `as_dict` actually returns list of dicts 🙂

KapJI: @leikoilja I still think this function is way too hacky and I'm going to remove it 🙂

DurgNomis-drol: haha 😆

leikoilja: haha, @KapJI, as long as we have a working way to return data to sensors since they require data as dict

PR #115, opened 2021-03-26
Add support for the status attribute of alarms/timers

AstroViking:
First of all, thank you guys for working on this. The integration works perfectly for me ❤️ 

I have implemented support for the status attribute (set/ringing/snoozed) for alarms/timers, as I have used this for automations with a different integration before.

From https://rithvikvibhu.github.io/GHLocalApi/#operation/GetAlarmsandTimers:

* Alarms: status is 1 for set up and 2 for ringing. (3 is snoozed, found that out by trial and error 😅 )
* Timers: status is 1 for set up and 3 for ringing.

I have also added the status of the next alarm as a first level state attribute "status" for easier access.
Let me know, what you think :)

### DIFF
--- a/custom_components/google_home/models.py
+++ b/custom_components/google_home/models.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from datetime import timedelta
+from enum import Enum
 from typing import List, Optional
 
 from homeassistant.util.dt import as_local, utc_from_timestamp
@@ -45,6 +46,7 @@ class GoogleHomeDevice:
             GoogleHomeAlarm(
                 alarm_id=alarm["id"],
                 fire_time=alarm["fire_time"],
+                status=alarm["status"],
                 label=alarm.get("label"),
                 recurrence=alarm.get("recurrence"),
             )
@@ -58,6 +60,7 @@ class GoogleHomeDevice:
                 timer_id=timer["id"],
                 fire_time=timer["fire_time"],
                 duration=timer["original_duration"],
+                status=timer["status"],
                 label=timer.get("label"),
             )
             for timer in timers
@@ -90,11 +93,13 @@ class GoogleHomeTimer:
         timer_id: str,
         fire_time: int,
         duration: int,
+        status: int,
         label: Optional[str],
     ) -> None:
         self.timer_id = timer_id
         self.duration = str(timedelta(seconds=convert_from_ms_to_s(duration)))
         self.fire_time = convert_from_ms_to_s(fire_time)
+        self.status = GoogleHomeTimerStatus(status)
         self.label = label
 
         dt_utc = utc_from_timestamp(self.fire_time)
@@ -108,6 +113,7 @@ class GoogleHomeTimer:
             "timer_id": self.timer_id,
             "fire_time": self.fire_time,
             "duration": self.duration,
+            "status": self.status.name.lower(),
             "label": self.label,
         }
 
@@ -119,12 +125,14 @@ class GoogleHomeAlarm:
         self,
         alarm_id: str,
         fire_time: int,
+        status: int,
         label: Optional[str],
         recurrence: Optional[str],
     ) -> None:
         self.alarm_id = alarm_id
         self.recurrence = recurrence
         self.fire_time = convert_from_ms_to_s(fire_time)
+        self.status = GoogleHomeAlarmStatus(status)
         self.label = label
 
         dt_utc = utc_from_timestamp(self.fire_time)
@@ -137,6 +145,24 @@ class GoogleHomeAlarm:
         return {
             "alarm_id": self.alarm_id,
             "fire_time": self.fire_time,
+            "status": self.status.name.lower(),
             "label": self.label,
             "recurrence": self.recurrence,
         }
+
+
+class GoogleHomeAlarmStatus(Enum):
+    """Definition of Google Home alarm status"""
+
+    NONE = 0
+    SET = 1
+    RINGING = 2
+    SNOOZED = 3
+
+
+class GoogleHomeTimerStatus(Enum):
+    """Definition of Google Home timer status"""
+
+    NONE = 0
+    SET = 1
+    RINGING = 3
--- a/custom_components/google_home/sensor.py
+++ b/custom_components/google_home/sensor.py
@@ -16,7 +16,13 @@ from .const import (
     LABEL_TIMERS,
 )
 from .entity import GoogleHomeBaseEntity
-from .models import GoogleHomeAlarmDict, GoogleHomeDevice, GoogleHomeTimerDict
+from .models import (
+    GoogleHomeAlarmDict,
+    GoogleHomeAlarmStatus,
+    GoogleHomeDevice,
+    GoogleHomeTimerDict,
+    GoogleHomeTimerStatus,
+)
 from .types import (
     AddEntitiesCallback,
     AlarmsAttributes,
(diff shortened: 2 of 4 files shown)


custom_components/google_home/const.py
```diff
@@ -57,6 +58,19 @@
...
+STATUS_TIMER_SET_INT = 1
+STATUS_TIMER_RINGING_INT = 3
+STATUS_ALARM_SET_INT = 1
+STATUS_ALARM_RINGING_INT = 2
+STATUS_ALARM_SNOOZED_INT = 3
+
+STATUS_NONE = "none"
+STATUS_SET = "set"
+STATUS_RINGING = "ringing"
+STATUS_SNOOZED = "snoozed"
```
leikoilja, line 73: Do you guys think we should better go with a dictionary mapping here?
I am thinking of something like:
```python
STATUS_SET_LABEL = "set"
STATUS_SET_INT = 1
...
STATUS_MAPPING = {
  LABEL_ALARMS: {
   STATUS_SET_INT: STATUS_SET_LABEL,
    ... 
  },  
  LABEL_TIMERS: {
  ...
  },
}
```
Then we can reuse the same `STATUS_SET_LABEL` and `STATUS_SET_INT` for both timers and alarms and maintain a bit cleaner version of mapping. 

Then we can probs get rid of introduced `def map_timer_status()` when we need to match received status integer to a label in favour of:
```python
STATUS_MAPPING[LABEL_TIMER].get(status, None)
```

custom_components/google_home/models.py
```diff
@@ -22,6 +32,25 @@ def convert_from_ms_to_s(timestamp: int) -> int:
...
+
+
+def map_alarm_status(status: int) -> str:
+        """Maps status int to string label"""
+        mapping_dict = {
+            STATUS_ALARM_SET_INT: STATUS_SET,
+            STATUS_ALARM_RINGING_INT: STATUS_RINGING,
+            STATUS_ALARM_SNOOZED_INT: STATUS_SNOOZED
+        }
+        return mapping_dict.get(status, STATUS_NONE)
```
leikoilja, line 51: Can possibly be simplified. See proposition above :)

leikoilja (commented):
Awesome job! 🚀 
Thank you for contributing, @AstroViking 💥 
Please see a tiny comment i have

We are using `pre-commit` hooks to run a bunch of linters and checks. Please see [CONTRIBUTION](https://github.com/leikoilja/ha-google-home/blob/master/CONTRIBUTING.md#use-a-consistent-coding-style) section to find out more about `poetry` and `pre-commit` and run `poetry run pre-commit run --all-files` to fix all the lining errors so the actions pass.

custom_components/google_home/const.py
KapJI: I think it's better to use Enum here.

custom_components/google_home/const.py
```diff
@@ -36,6 +36,7 @@
 LABEL_TIMERS = "timers"
 LABEL_DEVICE = "device"
 
+
```
KapJI, line 39: Remove.

custom_components/google_home/const.py
KapJI: Let's keep strings here but integer values inside `Enum` in `models.py` which maps these values to corresponding strings.

custom_components/google_home/models.py
```diff
@@ -90,34 +121,37 @@ def __init__(
...
         fire_time: int,
+        status: int,
         duration: int,
         label: Optional[str],
     ) -> None:
         self.timer_id = timer_id
         self.duration = str(timedelta(seconds=convert_from_ms_to_s(duration)))
         self.fire_time = convert_from_ms_to_s(fire_time)
         self.label = label
+        self.status = map_timer_status(status)
```
KapJI, line 132: Construct `Enum` here.

custom_components/google_home/sensor.py
```diff
@@ -135,6 +136,7 @@ def state(self) -> Optional[str]:
     def device_state_attributes(self):
         """Return the state attributes."""
         return {
+            "status": self._get_next_alarm_status(),
```
KapJI, line 139:
```suggestion
            "next_alarm_status": self._get_next_alarm_status(),
```

custom_components/google_home/sensor.py
```diff
@@ -143,6 +145,12 @@ def _get_alarms_data(self) -> List[Dict[Any, Any]]:
         """Update alarms data extracting it from coordinator"""
         device = self.get_device()
         return self.as_dict(device.get_sorted_alarms()) if device else []
+    
+    def _get_next_alarm_status(self) -> str:
+        """Update next alarm status from coordinator"""
+        device = self.get_device()
+        next_alarm = device.get_next_alarm()
```
KapJI, line 152: Return `STATUS_NONE` if `device is None`.

KapJI (changes requested):

custom_components/google_home/sensor.py
Drakulix: Could we differentiate between the device reporting *no status* and the device being unreachable?
E.g. If my network has a hiccup at night and the device is temporarily not available, I do not want my wake-up automation to go off. But otherwise triggering on a state change from "ringing" to "none" would be super convenient.

Maybe add an additional STATUS_UNAVAILABLE? Or just not update the status and cache it?

KapJI: This for status of individual alarm not state of entity. 

But right now we have an issue with using `unavailable` for state when there are no alarms set. This is because we need to use sensor device class `timestamp` which only support such state when no actual timestamp is set. This causes entities appearing unavailable on the list of entities defined by integration. I think we can override `is_available` method of sensor entity and set it manually.

custom_components/google_home/const.py
AstroViking: @KapJI 
Thank you for suggesting using an Enum. I was not sure what you meant with keeping strings in `const.py` as I am not sure how you could use an Enum as name for a constant. I now removed status definitions completely from the constants and simply used `GoogleHomeAlarmStatus(status).name.lower()` to get a string value again.

Are you fine with this?

AstroViking:
@leikoilja @KapJI 

Thank you for your suggestions. I should have read the contribution guide before ;-)

Sorry I had quite a busy week and did not get around to look at it again until today.
I actually started from scratch using the new typings @KapJI added.

custom_components/google_home/sensor.py
```diff
@@ -172,10 +189,21 @@ def state(self) -> Optional[str]:
...
             "timers": self._get_timers_data(),
             "integration": DOMAIN,
         }
 
+    def _get_next_timer_status(self) -> str:
+        """Update next timer status from coordinator"""
+        device = self.get_device()
+        next_timer = device.get_next_timer() if device else None
+        return (
+            next_timer.status.name.lower()
```
KapJI, line 202: This is actually even better than using const string 🙂

KapJI (changes requested):
Please update `README.md` and describe this feature there.

custom_components/google_home/const.py
KapJI: Yes, that's great, thank you!

AstroViking:
I updated the `README.md` to reflect the changes there. However the linting currently fails during installing dependencies. I actually have no idea why this happens as I haven't changed anything in this regard 🤔

KapJI:
Yeah, that's weird. Maybe something is wrong with PyPI.

KapJI (approved):
Your changes look good, we need to figure out why CI is broken.

AstroViking:
Yeah, just checked https://status.python.org/ and sure enough:

image

KapJI:
Should be resolved, you can try again.

leikoilja:
Awesome stuff! Thank you guys! 🚀

PR #117, opened 2021-03-26
Add more type hints and enforce more checks

KapJI:
Everything except `config_flow` is pretty strictly typed now.
Still missing home assistant typing stubs and getting rid of `Any`.

Tested locally that it still works.

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -1,14 +1,15 @@
 """Sample API Client."""
 from asyncio import gather
 import logging
-from typing import List
+from typing import List, Optional
 
 import aiohttp
-from glocaltokens.client import GLocalAuthenticationTokens
+from glocaltokens.client import Device, GLocalAuthenticationTokens
 from glocaltokens.utils.token import is_aas_et
 from zeroconf import Zeroconf
 
 from homeassistant.const import HTTP_NOT_FOUND, HTTP_OK, HTTP_UNAUTHORIZED
+from homeassistant.core import HomeAssistant
 
 from .const import (
     API_ENDPOINT_ALARMS,
@@ -31,13 +32,13 @@ class GlocaltokensApiClient:
 
     def __init__(
         self,
-        hass,
+        hass: HomeAssistant,
         username: str,
         password: str,
         session: aiohttp.ClientSession,
-        android_id: str = None,
-        zeroconf_instance: Zeroconf = None,
-    ) -> None:
+        android_id: Optional[str] = None,
+        zeroconf_instance: Optional[Zeroconf] = None,
+    ):
         """Sample API Client."""
         self.hass = hass
         self._username = username
@@ -54,10 +55,10 @@ class GlocaltokensApiClient:
         self.google_devices: List[GoogleHomeDevice] = []
         self.zeroconf_instance = zeroconf_instance
 
-    async def async_get_master_token(self):
+    async def async_get_master_token(self) -> str:
         """Get master API token"""
 
-        def _get_master_token():
+        def _get_master_token() -> Optional[str]:
             return self._client.get_master_token()
 
         master_token = await self.hass.async_add_executor_job(_get_master_token)
@@ -65,13 +66,13 @@ class GlocaltokensApiClient:
             raise InvalidMasterToken
         return master_token
 
-    async def get_google_devices(self):
+    async def get_google_devices(self) -> List[GoogleHomeDevice]:
         """Get google device authentication tokens.
         Note this method will fetch necessary access tokens if missing"""
 
         if not self.google_devices:
 
-            def _get_google_devices():
+            def _get_google_devices() -> List[Device]:
                 return self._client.get_google_devices(
                     zeroconf_instance=self.zeroconf_instance,
                     force_homegraph_reload=True,
@@ -89,16 +90,16 @@ class GlocaltokensApiClient:
             ]
         return self.google_devices
 
-    async def get_android_id(self):
+    async def get_android_id(self) -> Optional[str]:
         """Generate random android_id"""
 
-        def _get_android_id():
+        def _get_android_id() -> Optional[str]:
             return self._client.get_android_id()
 
         return await self.hass.async_add_executor_job(_get_android_id)
 
     @staticmethod
-    def create_url(ip_address, port, api_endpoint):
+    def create_url(ip_address: str, port: int, api_endpoint: str) -> str:
         """Creates url to endpoint.
         Note: port argument is unused because all request must be done to 8443"""
         url = "https://{ip_address}:{port}/{endpoint}".format(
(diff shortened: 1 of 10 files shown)


custom_components/google_home/api.py
```diff
@@ -198,7 +201,7 @@ async def update_google_devices_information(self):
 
         coordinator_data = await gather(
             *[
-                self.get_alarms_and_timers(device)
+                self.get_alarms_and_timers(device, device.ip_address)
```
leikoilja, line 204: Why explicitly passing in `ip_address`?

custom_components/google_home/models.py
```diff
@@ -40,26 +51,26 @@ def __init__(
...
         self._alarms = [
             GoogleHomeAlarm(
-                alarm_id=alarm[ID],
-                fire_time=alarm[FIRE_TIME],
-                label=alarm.get(LABEL),
-                recurrence=alarm.get(RECURRENCE),
+                alarm_id=alarm["id"],
+                fire_time=alarm["fire_time"],
+                label=alarm.get("label"),
+                recurrence=alarm.get("recurrence"),
```
leikoilja, line 61: Why are we back at hardcoding the values? What if one day google device to change a parameter name from `fire_time` to `fire_time_ms` or something else? We would need to go through multiple places with hardcoded `fire_time` to manually replace those? Is it not better to have one place(as const) where all these hardcoded values are stored and maintained? 🤔

custom_components/google_home/models.py
```diff
@@ -83,6 +94,15 @@ def get_next_timer(self) -> Optional[GoogleHomeTimer]:
         return timers[0] if timers else None
 
 
+class GoogleHomeTimerDict(TypedDict):
+    """Typed dict representation of Google Home time"""
+
+    timer_id: str
+    fire_time: int
+    duration: str
+    label: Optional[str]
```
leikoilja, line 103: I suggest we move out `TypedDict` to a separate file like `types.py`. It would give it a bit better structure

custom_components/google_home/models.py
```diff
@@ -3,25 +3,36 @@
...
+    recurrence: Optional[str]
+
+
+class TimerJsonDict(TypedDict, total=False):
+    """Typed dict for JSON representation of timer returned by Google Home API"""
+
+    id: str
+    fire_time: int
+    original_duration: int
+    label: Optional[str]
```
leikoilja, line 33: I suggest we move out `TypedDict` to a separate file like `types.py`. They are not "core" relevant, thou important to have. It would give it a bit better structure

custom_components/google_home/sensor.py
```diff
@@ -60,6 +61,17 @@ async def async_setup_entry(
...
 
+class DeviceAttributes(TypedDict):
+    """Typed dict for device attributes"""
+
+    device_name: str
+    auth_token: Optional[str]
+    ip_address: Optional[str]
+    hardware: Optional[str]
+    available: bool
+    integration: str
```
leikoilja, line 72: I suggest we move out `TypedDict` to a separate file like `types.py`. It would give it a bit better structure

custom_components/google_home/sensor.py
```diff
@@ -105,6 +117,13 @@ def get_device_attributes(device):
         }
 
 
+class AlarmsAttributes(TypedDict):
+    """Typed dict for alarms attributes"""
+
+    alarms: List[GoogleHomeAlarmDict]
+    integration: str
```
leikoilja, line 124: I suggest we move out `TypedDict` to a separate file like `types.py`. It would give it a bit better structure

custom_components/google_home/sensor.py
```diff
@@ -132,17 +151,26 @@ def state(self) -> Optional[str]:
...
+        return (
+            [alarm.as_dict() for alarm in device.get_sorted_alarms()] if device else []
+        )
+
+
+class TimersAttributes(TypedDict):
+    """Typed dict for timers attributes"""
+
+    timers: List[GoogleHomeTimerDict]
+    integration: str
```
leikoilja, line 173: I suggest we move out `TypedDict` to a separate file like `types.py`. It would give it a bit better structure

custom_components/google_home/sensor.py
```diff
@@ -172,14 +200,16 @@ def state(self) -> Optional[str]:
...
         }
 
-    def _get_timers_data(self) -> List[Dict[Any, Any]]:
+    def _get_timers_data(self) -> List[GoogleHomeTimerDict]:
         """Update timers data extracting it from coordinator"""
         device = self.get_device()
-        return self.as_dict(device.get_sorted_timers()) if device else []
+        return (
+            [timer.as_dict() for timer in device.get_sorted_timers()] if device else []
+        )
```
leikoilja, line 215: Nice fix! 💥

leikoilja (approved):
In general, look good! Thanks for bringing us here - to a pretty strictly typed project. Appreciate it! 💥 

Have a few comments thou, but they are IMHO.

custom_components/google_home/api.py
KapJI: Because `device.ip_address` is `Optional[str]` although we checked here that it's not `None`. In this function we will need `str` to construct url so this is linting time check for that.

Another option is explicitly checking again in `get_alarms_and_timers` that `device.ip_address` is not `None` but that will be runtime check.

custom_components/google_home/models.py
KapJI: `TypedDict` only allows string literals here. But it verifies this strings match fields in corresponding `TypedDict`.

custom_components/google_home/api.py
leikoilja: Can we just remove `Optional` from `GoogleHomeDevice`'s `ip_address` field and make it mandatory with a fallback to `None`?

KapJI: `Optional` is returned from `glocaltokens`. I think there can be a case when ip address is not found in Foyer API.

custom_components/google_home/types.py
```diff
@@ -0,0 +1,84 @@
...
+from homeassistant.helpers.entity import Entity
+
+
+class AddEntitiesCallback(Protocol):
+    """Protocol type for async_setup_entry callback"""
+
+    def __call__(
+        self, new_entities: Iterable[Entity], update_before_add: bool = False
+    ) -> None:
+        ...
```
leikoilja, line 15: Fancy stuff. And i believe we'll possibly be able to use that from HA stubs (weather they official ones or 3rd party), right? :)

custom_components/google_home/types.py
```diff
@@ -0,0 +1,84 @@
...
+    hardware: Optional[str]
+    available: bool
+    integration: str
+
+
+class AlarmsAttributes(TypedDict):
+    """Typed dict for alarms attributes"""
+
+    alarms: List[GoogleHomeAlarmDict]
+    integration: str
```
leikoilja, line 69: Being picky here, shall we order these type classes alphabetically? 👍

leikoilja (approved):
Superb! Thank you 💥

custom_components/google_home/types.py
KapJI: Yes, I'll be working on them soon.

custom_components/google_home/types.py
KapJI: Some of them depend on each other, so you can't use arbitrary order. I tried to group them by order of appearence.

PR #131, opened 2021-04-01
Add reuse of master_token

### LINKED ISSUE #130: Reuse master_token when the integration is reloaded.
## Describe the bug

Reuse `master_token` when HA is restarted and the integration i reloaded. Right now we save the `master_token` in `ConfigEntry`, but we don't reuse it. This will fix issues where you can get logged out of your Google Account on other devices.

## Version of the integration

```text
Latest
```

## Glocaltokens Version (Underlying package)

```text
Not relevant
```

## Logs

```text
Not relevant
```

## Debug log

```yaml
logger:
  default: debug
  logs:
    custom_components.google_home: debug
```

```text

Not relevant

```

DurgNomis-drol:
Small changes to how we login.

We now use `master_token` instead of username when the integration is initiated in `__init_.py`.

This fixes random logouts, and it reuses the master_token instead of requesting it every time the integration is reloaded. At restart or manually. 

I have tested it locally and it works when I hit reload integration.

This closes  #130 and solves this discussion #126

### DIFF
--- a/custom_components/google_home/__init__.py
+++ b/custom_components/google_home/__init__.py
@@ -10,6 +10,7 @@ import logging
 
 from homeassistant.components import zeroconf
 from homeassistant.config_entries import ConfigEntry
+from homeassistant.const import CONF_PASSWORD, CONF_USERNAME
 from homeassistant.core import Config, HomeAssistant
 from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
@@ -40,6 +41,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         hass.data.setdefault(DOMAIN, {})
         _LOGGER.info(STARTUP_MESSAGE)
 
+    username = entry.data.get(CONF_USERNAME)
+    password = entry.data.get(CONF_PASSWORD)
     android_id = entry.data.get(CONF_ANDROID_ID)
     master_token = entry.data.get(CONF_MASTER_TOKEN)
 
@@ -47,8 +50,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     zeroconf_instance = await zeroconf.async_get_instance(hass)
     glocaltokens_client = GlocaltokensApiClient(
-        hass,
-        session,
+        hass=hass,
+        session=session,
+        username=username,
+        password=password,
         master_token=master_token,
         android_id=android_id,
         zeroconf_instance=zeroconf_instance,
--- a/custom_components/google_home/__init__.py
+++ b/custom_components/google_home/__init__.py
@@ -18,8 +18,7 @@ from homeassistant.helpers.update_coordinator import DataUpdateCoordinator
 from .api import GlocaltokensApiClient
 from .const import (
     CONF_ANDROID_ID,
-    CONF_PASSWORD,
-    CONF_USERNAME,
+    CONF_MASTER_TOKEN,
     DOMAIN,
     PLATFORMS,
     SENSOR,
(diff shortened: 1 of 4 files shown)


custom_components/google_home/config_flow.py
```diff
@@ -49,10 +49,10 @@ async def async_step_user(self, user_input=None):
             username = user_input[CONF_USERNAME]
             session = async_create_clientsession(self.hass)
             client = GlocaltokensApiClient(
-                hass=self.hass,
+                self.hass,
+                session,
                 username=user_input[CONF_USERNAME],
                 password=user_input[CONF_PASSWORD],
```
leikoilja, line 55: I think we should keep all arguments as `kwargs` keeping `hass=...` and `session=...`, I know it's not mandatory, but it's a good practice and is easier to follow the code :)

custom_components/google_home/__init__.py
```diff
@@ -41,15 +40,18 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
 
     zeroconf_instance = await zeroconf.async_get_instance(hass)
     glocaltokens_client = GlocaltokensApiClient(
-        hass, username, password, session, android_id, zeroconf_instance
+        hass,
+        session,
+        master_token=master_token,
+        android_id=android_id,
+        zeroconf_instance=zeroconf_instance,
     )
```
leikoilja, line 55: I really think we should have a fallback to `username` and `password` if `master_token` is not there.
We should provide "all we have" and let `glocaltokens` package decide what should be used. 

We can safely keep `username`, `password` and add `master_token` here, because looking at how `glocaltokens` client is working now it will not use `username` and `password` if `master_token` was provided when initializing `GLocalAuthenticationTokens` client

custom_components/google_home/api.py
```diff
@@ -49,6 +50,7 @@ def __init__(
         self._client = GLocalAuthenticationTokens(
             username=username,
             password=password,
+            master_token=master_token,
```
KapJI, line 53: What will happen if both `username/password` and `master_token` are provided?

KapJI (approved):
LGTM, thanks!

custom_components/google_home/__init__.py
leikoilja: It would also be very nice to make all arguments `kwargs`

leikoilja (changes requested):
Great stuff! Thanks for nailing it! 

Just a few minor comments.
Also, please don't commit `poetry.lock` file. I believe there are a bunch of changes from your previous testings now. We should only change `poetry.lock` when the development environment is changed and we want to propagate these changes to the rest of developers (it will most likely be in a separate PR)

KapJI:
`poetry.lock` changes look like result of `poetry update`. Shouldn't be harmful.

custom_components/google_home/config_flow.py
KapJI: I think `mypy` guarantees order of arguments is good here so I think this is optional.

Btw, it's may be interesting. Python 3.8 introduced positional only arguments in PEP 570: https://www.python.org/dev/peps/pep-0570/

DurgNomis-drol:
I have no idea why `poetry.lock` is updated. But yes i did `poetry update` to make sure i was up to date with everything.

How do i omit `poetry.lock` with out adding it to `.gitignore` ?

See my latest push 😄

KapJI:
To update environment to current versions in `poetry.lock`. you need to use `poetry install` or `poetry install --remove-untracked`.

`poetry update` is for updating version in `poetry.lock`.

KapJI:
To restore file to previous version: `git checkout master -- poetry.lock`

DurgNomis-drol:
@KapJI Thanks for the clarification. So poetry update is not something I should use unless i add packages to poetry?

KapJI:
`poetry update` is for bumping package versions in `poetry.lock`. I think we need to configure `Dependabot` to do that.

DurgNomis-drol:
@KapJI Thanks, and yes that would be neat. 🎉 💥

leikoilja (approved):
Magical! Thank you so much! 💥

PR #132, opened 2021-04-02
Bump up dependancy versions + dependabot

leikoilja:
- Running `poetry update` to update package versions
- Tweaking dependabot config file to manage dependencies from now on


KapJI (changes requested):
Let's revert poetry.lock to see if Dependabot will really update it today or tomorrow.

KapJI (approved):
Thanks! 🎉

Issue #128, opened 2021-03-31
Don't produce errors when device is unreachable


KapJI:
## Describe the bug

When device is turned off the component shouldn't spam with errors like this, we can add a message on `debug` level about this. But authentication errors (when token is invalid) are still worth an error message.

## Logs

`2021-03-31 10:59:33 ERROR (MainThread) [custom_components.google_home] Request error: Cannot connect to host 192.168.XXX.XX:8443 ssl:default [None]`

```
Logger: custom_components.google_home
Source: custom_components/google_home/api.py:162
Integration: Google Home (documentation, issues)
First occurred: 30 maart 2021 9:03:53 (4754 occurrences)
Last logged: 11:18:53

Request error: Cannot connect to host 192.168.XXX.XX:8443 ssl:default [Connect call failed ('192.168.XXX.X', 8443)]
Request error: Cannot connect to host 192.168.XXX.XX:8443 ssl:default [Connect call failed ('192.168.XXX.XX', 8443)]
Request error: [Errno 104] Connection reset by peer
Request error: Cannot connect to host 192.168.XXX.XX:8443 ssl:default [None]
Request error: Cannot connect to host 192.168.XXX.XX:8443 ssl:default [Connect call failed ('192.168.XXX.XX', 8443)]
```

Issue #122, opened 2021-03-27
Introduce periodic Foyer device fetching


leikoilja:
**Is your feature request related to a problem? Please describe.**
We need to fetch Foyer devices every 10 mins instead of 24h as it is right now. The problem is when a new device is plugged in the user expects the device to appear in `Google Home` integration, however, with a current setup, we only fetch Foyer devices every time the access token goes off (~24h). 

**Describe the solution you'd like**
Introduce a constant parameter to adjust how often the integration refreshes Foyer devices (does a `force` fetching of all google devices when calling `glocaltokens.get_devices`).

leikoilja:
@KapJI, @Drakulix, @ArnyminerZ, @DurgNomis-drol anyone up for a fun task? 😅

PR #133, opened 2021-04-02
Retry device update on every coordinator update

### LINKED ISSUE #119: Retry alarms/timers fetching from every discovered device on every coordinator update
## Describe the bug
Google Home devices are mistakenly filtered:
Devices that actually contain alarms/timers are filtered out if the API request fails to establish because of underlying issues of `pychromcast`.
Please see there two log entries where my Google Home devices were filtered out because the API request to them unexpectedly returned `404`.

## Version of the integration

```text
Google Home v1.0.1
```

## Logs
Screenshot 2021-03-27 at 08 32 23
Screenshot 2021-03-27 at 08 33 33

From these logs, you see that I used to have a working `Kitchen` device that got filtered out and no more data fetched from it (1st screenshot) and `Living Room` device that also got filtered out (2nd screenshot)

## Expected behaviour
Retry fetching alarms/timers from device(s) on every coordinator update

leikoilja:
Closes #119 

- I have spent some time looking for a more clever way to differentiate if the device has timers/alarms and might found something interesting. When calling `eureka_info` endpoint ([ref](https://rithvikvibhu.github.io/GHLocalApi/#operation/EurekaInfo)) we can probably look at `device.device_info. capabilities` for a presence of key `assistant_supported=True`.  I am not 100% `assistant_supported` will correlate to supporting alarms/timers, but it's worth a shoot to test with the variety of devices we have on our hands. If it will we can probably introduce better logic at sorting devices when `get_google_device` with an extra call to `eureka_info`?

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -130,6 +130,21 @@ class GlocaltokensApiClient:
             ) as response:
                 if response.status == HTTP_OK:
                     resp = await response.json()
+                    device.available = True
+                    if resp:
+                        if JSON_TIMER in resp or JSON_ALARM in resp:
+                            device.set_timers(resp.get(JSON_TIMER))
+                            device.set_alarms(resp.get(JSON_ALARM))
+                        else:
+                            _LOGGER.error(
+                                (
+                                    "Failed to parse fetched data for device %s - %s. "
+                                    "Received = %s"
+                                ),
+                                device.name,
+                                API_RETURNED_UNKNOWN,
+                                resp,
+                            )
                 elif response.status == HTTP_UNAUTHORIZED:
                     # If token is invalid - force reload homegraph providing new token
                     # and rerun the task.
@@ -142,18 +157,20 @@ class GlocaltokensApiClient:
                     )
                     # We need to retry the update task instead of just cleaning the list
                     self.google_devices = []
-                elif response.status == HTTP_NOT_FOUND:
                     device.available = False
+                elif response.status == HTTP_NOT_FOUND:
                     _LOGGER.debug(
                         (
                             "Failed to fetch data from %s, API returned %d. "
-                            "The device(hardware='%s') is not Google Home "
-                            "compatable and has no alarms/timers."
+                            "The device(hardware='%s') is possibly not Google Home "
+                            "compatable and has no alarms/timers. "
+                            "Will retry later."
                         ),
                         device.name,
                         response.status,
                         device.hardware,
                     )
+                    device.available = False
                 else:
                     _LOGGER.error(
                         "Failed to fetch %s data, API returned %d: %s",
@@ -161,8 +178,8 @@ class GlocaltokensApiClient:
                         response.status,
                         response,
                     )
+                    device.available = False
         except ClientConnectorError:
-            # Check if this is where we are getting with incorrect token
             _LOGGER.debug(
                 (
                     "Failed to connect to %s device. "
@@ -170,6 +187,7 @@ class GlocaltokensApiClient:
                 ),
                 device.name,
             )
+            device.available = False
         except ClientError as ex:
             # Make sure that we log the exception if one occurred.
             # The only reason we do this broad is so we easily can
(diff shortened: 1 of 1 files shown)


custom_components/google_home/sensor.py
```diff
@@ -40,17 +40,21 @@ async def async_setup_entry(
...
-                    coordinator,
-                    device.name,
-                ),
-                GoogleHomeTimersSensor(
-                    coordinator,
-                    device.name,
-                ),
-            ]
+        if device.auth_token:
+            if device.has_alarms():
```
KapJI, line 44: This looks like that sensor won't be created if device doesn't have any alarms *set*.
What will be `self._alarms` value when device supports alarms but they're not set?

leikoilja: Yeh, that what i meant in descriptions:

> The behavior is not ideal still. In sensor.py under async_setup_entry(once during HA bootup) we setup sensors if the device has token and timers/alarms, but what happens to devices that are not yet available(like offline), those will be forever skipped, since no sensors will be created for those devices until HA is restarted again. If that's true - it defeats the purpose of #122

`self._alarms` is just an empty list if no alarms were fetched during api call, right ?

leikoilja:
@KapJI more thoughts on that one? 🤔

KapJI:
@leikoilja I don't get how this is supposed to work. If you don't have any alarms or timers set, then you won't have any sensors created. Although you will query the API, this won't create any sensor entities.

custom_components/google_home/models.py
```diff
@@ -35,10 +35,17 @@ def __init__(
         self.auth_token = auth_token
         self.ip_address = ip_address
         self.hardware = hardware
-        self.available = True
         self._timers: List[GoogleHomeTimer] = []
         self._alarms: List[GoogleHomeAlarm] = []
 
+    def has_alarms(self) -> bool:
+        """Returns whether or not device has alarms"""
+        return self._alarms != []
```
KapJI, line 43: I'd rather use this:
```suggestion
        return len(self._alarms) > 0
```

leikoilja:
yeh, exactly. Well, that would be a problem for only those devices that are connected to one's google home account but are not online during HA startup. 
Then during HA startup no alarms/timers will be fetched and therefore no corresponding sensors created.
If we just keep it as currently is:
`if device.auth_token: # create sensor entities` then we'd end up with a bunch of spam sensors like timer/alarm sensor for devices like `chromecast`, `shield` etc

KapJI:
As I understand it, it's worse than that. It's not if device is offline during HA startup or device doesn't support alarms then sensors won't be created. It's if there're not alarms **set** during the startup then sensors won't be created. Correct me if I'm wrong.

leikoilja:
uuuuu, you're absolutely right, I completely overlooked it... 😅 
Well I guess we drop those check `has_timers` and `has_alarms` and just "spam" create timer/alarm sensors for all devices we discover (incl shield, chromecast) for now then? I don't see a good way we could "ignore" certain devices without maintaining black/white lists or producing false positives like in the bug we're trying to solve here #119

KapJI:
I think this is how it should work:
- During setup we create sensors for all devices which correctly respond on alarms endpoint even if they return empty list.
- All devices returned by Foyed API are being queried in local API calls. `Bad Authentication` is a valid error here, `Host Unreachable` is only a debug level message. If we can detect that auth is correct but device doesn't have such endpoint or doesn't have some capability, we can stop querying it.
- If device starts to repond on alarms endpoint we can create corresponding entities for it dynamically, I think HA supports this.

leikoilja:
That's some good thinking right there.

So, as for the:
1. I think i can fix it
2. We already query Foyer API devices in `self.google_devices` on the client; `Bad Authentication` is currently perceived as an invalid token, meaning it's expired, so what we do is - we clean `self.google_devices` and refetch Foyer API devices with new tokens. `Host Unreachable` was solved in #128. Regarding detection if device does not support endpoint by looking at capabilities from `eureka_info` we can discuss in a separate issue, i think it could work.
3. Let's do dynamic sensor creation (if the device started responding) in a separate issue, cause it seems to be a big one?

custom_components/google_home/api.py
```diff
@@ -178,17 +196,7 @@ async def get_alarms_and_timers(
                 "Request error: %s",
                 ex,
             )
-        finally:
```
leikoilja, line 181: I believe this `finally` is actually redundant because:

> The finally block lets you execute code, regardless of the result of the try- and except blocks.

Therefore, i have just moved the response parsing close to where `resp` is obtained

KapJI (approved):
LGTM 🚀

PR #138, opened 2021-04-02
Adding ClientConnectionError handler

### LINKED ISSUE #128: Don't produce errors when device is unreachable

leikoilja:
Closes #128,

We should not flood HA logs with errors if we can't connect to the device (happens when some devices go offline). Therefore let's catch the ConnectionError and log as `debug`.

I have tested locally by fetching data from device and then unplugging it

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -143,6 +143,7 @@ class GlocaltokensApiClient:
                     # We need to retry the update task instead of just cleaning the list
                     self.google_devices = []
                 elif response.status == HTTP_NOT_FOUND:
+                    device.available = False
                     _LOGGER.debug(
                         (
                             "Failed to fetch data from %s, API returned %d. "
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -147,9 +147,8 @@ class GlocaltokensApiClient:
                     _LOGGER.debug(
                         (
                             "Failed to fetch data from %s, API returned %d. "
-                            "The device(hardware='%s') is possibly not Google Home "
-                            "compatable and has no alarms/timers. "
-                            "Will retry later."
+                            "The device(hardware='%s') is not Google Home "
+                            "compatable and has no alarms/timers."
                         ),
                         device.name,
                         response.status,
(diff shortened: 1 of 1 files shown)


custom_components/google_home/api.py
```diff
@@ -142,12 +143,12 @@ async def get_alarms_and_timers(
                     # We need to retry the update task instead of just cleaning the list
                     self.google_devices = []
                 elif response.status == HTTP_NOT_FOUND:
-                    device.available = False
```
leikoilja, line 145: flip, that's from a different PR. Will remove rn

KapJI (approved):
Looks good, thanks! 🎉

custom_components/google_home/api.py
```diff
@@ -146,8 +147,9 @@ async def get_alarms_and_timers(
                     _LOGGER.debug(
                         (
                             "Failed to fetch data from %s, API returned %d. "
-                            "The device(hardware='%s') is not Google Home "
-                            "compatable and has no alarms/timers."
+                            "The device(hardware='%s') is possibly not Google Home "
+                            "compatable and has no alarms/timers. "
+                            "Will retry later."
```
KapJI, line 152: This also looks unrelated to this PR.

leikoilja: Ops, you re right, thanks

PR #139, opened 2021-04-02
Move version badge to the beginning of the list

KapJI:
- I think version should be closer to the front of this list as it's more relevant.
- Change HACS Custom to HACS Default as the integration is now available in the default store.
- Btw, should we remove some of these badges? License and number of contributors are already displayed on the github main page. Right now it feels a bit cluttered and they don't fit in one line.
- Should we use more compact default style badges?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,7 @@
 [![GitHub Workflow Status][workflow-shield]][workflow]
 [![GitHub Release][releases-shield]][releases]
 [![hacs][hacsbadge]][hacs]
-[![Contributors][contributors-shield]][contributors]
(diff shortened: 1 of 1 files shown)


leikoilja (commented):
I like all the proposed comments you have, it would be nice to just make those changes in the same PR :)

DurgNomis-drol:
I agree with all except the last one, I think the size of badges we use now is fine. 😄

KapJI:
Alright, so let's remove license and number of contributors badges but keep the same style.

leikoilja (approved):

DurgNomis-drol (approved):
LGTM 😄 

Great job!

PR #113, opened 2021-03-25
Readme: add troubleshooting section

brianhanifin:
New troubleshooting section includes:

* Collecting useful log data
* "Username/Password is incorrect"
* 2 Factor Authentication / App Passwords

Even though I haven't be able to login yet, I wanted to contribute a little bit of the troubleshooting knowledge I have used.

Issue #58, opened 2021-03-05
Add refresh setting to configuration 


leikoilja:
**Is your feature request related to a problem? Please describe.**
Currently the integration updates timers/alarms at a hardcoded period (15 seconds or 5 minutes). 

**Describe the solution you'd like**
A configuration UI setting where the user can set a custom update rate. Maybe with some reasonable limits.

KapJI:
I don't think we actually need this. 10 seconds is reasonable interval for everyone.

leikoilja:
Perhaps you are right :) We can revisit it if that will be needed, but as you say 10 sec is most likely gonna do great for most! :) Thanks!

PR #141, opened 2021-04-02
Adding config options 

### LINKED ISSUE #124: More options for config flow
**Is your feature request related to a problem? Please describe.**
We currently only ask for login and password on config flow. However some people might have issue with `BadAuthentication` cause by different custom components as #95 for example. As an alternative solution we could provide `master_token` input on config UI.
Also we [discuss](https://github.com/leikoilja/ha-google-home/discussions/116) to have error catching for which we need to provide an option to 'opt out' from diagnostic data collection.

**Describe the solution you'd like**
More versatile config flow UI that allows to enter master token and opt in / opt out from diagnotic data collection

leikoilja:
Closes #124

Adding options (when the integration added, there is the 'Options' button right next to 'Remove):
- Update interval 
- Allowing to collect diagnostics data

### DIFF
--- a/custom_components/google_home/translations/en.json
+++ b/custom_components/google_home/translations/en.json
@@ -11,7 +11,7 @@
       }
     },
     "error": {
-      "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords). If authentication still failes please see https://github.com/leikoilja/ha-google-home/issues/95"
+      "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords). If authentication still fails please see https://github.com/leikoilja/ha-google-home/issues/95"
     },
     "abort": {
       "single_instance_allowed": "Only a single instance is allowed."
--- a/custom_components/google_home/translations/en.json
+++ b/custom_components/google_home/translations/en.json
@@ -21,8 +21,7 @@
     "step": {
       "user": {
         "data": {
-          "data_collection": "Allow anonymous diagnostic data collection (See https://github.com/leikoilja/ha-google-home#diagnostic-data-collection). Restart required.",
-          "update_interval": "Update interval how often to fetch google device information (sec). Restart required."
+          "data_collection": "Allow anonymous diagnostic data collection (See https://github.com/leikoilja/ha-google-home#diagnostic-data-collection). Restart required."
         }
       }
     }
--- a/custom_components/google_home/translations/en.json
+++ b/custom_components/google_home/translations/en.json
@@ -6,12 +6,13 @@
         "description": "If you need help with the configuration have a look here: https://github.com/leikoilja/ha-google-home. Use your google account username and app password. It's safer/easier to generate an app password and use it instead of the actual password. It still has the same access as the regular password, but still better than using the real password while scripting. (https://myaccount.google.com/apppasswords). If not, regular google account password should work.",
         "data": {
           "username": "Google account username",
-          "password": "Google account app password"
+          "password": "Google account app password",
+          "data_collection": "Allow anonymous diagnostic data collection (See https://github.com/leikoilja/ha-google-home#diagnostic-data-collection)"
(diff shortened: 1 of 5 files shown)


KapJI:
I still don't think that we should make update interval configurable. It's an implementation detail which shouldn't be exposed to users. You may see that most components in HA Core don't allow to do that. At least we need some clear demand for this.

custom_components/google_home/translations/en.json
```diff
@@ -11,7 +11,7 @@
       }
     },
     "error": {
-      "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords)"
+      "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords). If authentication still failes please see https://github.com/leikoilja/ha-google-home/issues/95"
```
KapJI, line 14: I think we should use a link to a troubleshooting section. Do you want to complete #113?
```suggestion
      "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords). If authentication still fails please see https://github.com/leikoilja/ha-google-home/issues/95"
```

leikoilja:
Haven't we spoken about it? 🤔 
I believe @DurgNomis-drol wanted some better tuning of update interval. In my setup, i ll defs set the update interval to something of like 30secounds, maybe even a minute. There is no need to do as frequent updates as 10sec for my ~7 home devices that i use. All i do is "Set timer for 30 mins for laundry", so there is no fine resolution needed. 
I don't really see how it can hurt having it configurable. Most other integrations i have used (when they had `yaml` configurations) usually had interval configurable

custom_components/google_home/translations/en.json
leikoilja: Oh thanks 💥

KapJI:
Do you have any real issues with low update interval? These local queries should be super fast and we can actually query local devices even every second.

DurgNomis-drol:
I use timers with short durations sometimes, and i am very happy with 10 secs update interval. @KapJI  is right about that we should not expose low-level things like this to the user but then again, as @leikoilja mentions, it is an option many things have in `yaml`.

KapJI:
And I was under impression that we agreed to not add this in #58.

KapJI:
So I think unless there're no issues with current update interval, we should keep as is. If there will be any issues we should first try to solve them and then tune update interval or make it configurable.

KapJI:
The same idea applies to using `master_token` for login. We should fix `gpsoauth` to be more robust rather than doing this.

leikoilja:
yeh, i guess it's just me who really wants the update interval to be configurable with not that much of objective reasoning 😅 apparently as much that i have even forgotten you've already convinced me to drop that idea 😄  will revert back :)

custom_components/google_home/translations/en.json
```diff
@@ -11,7 +11,7 @@
       }
     },
     "error": {
-      "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords)"
+      "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords). If authentication still fails please see https://github.com/leikoilja/ha-google-home/issues/95"
```
KapJI, line 14: I still think that we need to complete Troubleshooting section in readme instead of this link 🙂

KapJI (approved):

custom_components/google_home/translations/en.json
leikoilja: Good point. Let me change the link to refer to README and then we can be more elaborate in Troubleshooting section

custom_components/google_home/translations/en.json
leikoilja: Sorry, @KapJI, haven't seen your comment. Will check #113 and change the link here to point to troubleshooting section

PR #145, opened 2021-04-02
Update release-drafter.yml

KapJI:
Move bug fixes to the top. They are usually more important.


leikoilja (approved):
Good point, haven't even though about setting an order here

KapJI:
Also moved `performance` to features 🙂

leikoilja:
is performance a feature? 🤔  quite philosophical question i guess

KapJI:
I'd say it's closer to feature than to anything else.
Something like `it's 2x faster` worth putting in the features list 🙂

PR #147, opened 2021-04-03
Fix Labeler

KapJI:
- https://github.com/crazy-max/ghaction-github-labeler keeps labels from `labels.yml` in sync with Github. It should run on push.
- https://github.com/actions/labeler assigns labels based on `labeler.yml` config. It wasn't running.
- Updated `labels.yml`.
- Added few more rules in `labeler.yml`.
- Remove `skip-delete: true`. New labels should be added in `labels.yml`.
- Removed `dev` branch from Github Actions because it doesn't exist.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,8 @@
-test:
-  - tests/*
+dependencies:
+  - poetry.lock
 
-github_structure:
+documentation:
+  - "*.md"
+
+github structure:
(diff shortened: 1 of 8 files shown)


leikoilja (approved):
Great stuff. I was wondering why labels were not automatically updated by `labeler` and needed some time to look into that. Thanks for fixing it, it's amazing 💣 💥

Issue #85, opened 2021-03-11
Enable strict typing


KapJI:
# Problem
Having types and type checks helps to:
- Check that all functions exist and called with correct arguments.
- Make the code more robust.
- Allow to refactor code simpler and safer.
# The plan
I'll enable `mypy` checks soon and will gradually enable stricter checks as more parts are covered with types.

Let me know what do you think!

leikoilja:
I think that would be a great additive! I have already set up my `vim` to support `mypy`, though haven't really adapted to using it just yet! 😄  But i think it would be great adding it

KapJI:
First step: #88

KapJI:
Opened a PR in homeassistant core: https://github.com/home-assistant/core/pull/47796

leikoilja:
Did they close home-assistant/core#47796 because they don't support typing across the entire `core` repo? 🤔  😞

KapJI:
Yeah, basically they have many untyped components (the actual core is strictly typed) and they fear enabling this can break typing of custom components (only those which import *untyped* standard components, popular ones like `sensor` or `light` are typed) therefore it's better to not mark their package as typed.

I don't agree with this but I actually have a workaround. I can generate typing stubs and publish them as a separate package which developer or custom components can install have type check their code against `homeassistant` package.

leikoilja:
uuu, that would be interesting to see :) 
So why is all that hassle? To be able to use typing hints, right? :D

KapJI:
Well, I think this should be useful. Not just for this project but other custom components as well.

leikoilja:
ou no no, i totally think it would be useful what you are proposing with making a stub as a separate package. 

I just mean the entire `mypy` is used for typing help, right? (as u probs already understood, that's quite a new concept for me at this point)

KapJI:
Yes, `mypy` is used for type checking. Without it (or another type checker like `pyre`) type hints are useless 🙂

KapJI:
As I mentioned in the description, having type hints helps to make sure you're passing arguments with the correct types and this helps to avoid mistakes.

leikoilja:
@KapJI, is there anything left from `typing` to implement?

KapJI:
Yes, few things.

1. Add `homeassistant-stubs`. I already created [the package](https://pypi.org/project/homeassistant-stubs/). Waiting for the release of homeassistant 2021.4.0 next Wednesday. I can probably start using beta version in the meantime.
[...]

PR #148, opened 2021-04-03
Use homeassistant typing stubs

KapJI:
This is based on beta version of Home Assistant: `2021.4.0b3`. I'll update it when the final version will be released on Wednesday.

This helped to find few bugs:
- The main one is that `async_create_task` should be used instead of `async_add_job`.
- Also `self._client.get_master_token()` may return `Optional[str]`. This wasn't processed correctly.

Part of #85

### DIFF
--- a/custom_components/google_home/__init__.py
+++ b/custom_components/google_home/__init__.py
@@ -11,7 +11,7 @@ import logging
 from homeassistant.components import zeroconf
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_PASSWORD, CONF_USERNAME
-from homeassistant.core import Config, HomeAssistant
+from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator
@@ -30,7 +30,7 @@ from .const import (
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
 
-async def async_setup(_hass: HomeAssistant, _config: Config) -> bool:
+async def async_setup(_hass: HomeAssistant, _config: dict) -> bool:
     """Set up this integration using YAML is not supported."""
     return True
 
@@ -76,7 +76,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     # Offload the loading of entities to the platform
     for platform in PLATFORMS:
-        hass.async_add_job(
+        hass.async_create_task(
             hass.config_entries.async_forward_entry_setup(entry, platform)
         )
 
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -61,8 +61,8 @@ class GlocaltokensApiClient:
     async def async_get_master_token(self) -> str:
         """Get master API token"""
 
-        def _get_master_token() -> Optional[str]:
-            return self._client.get_master_token()
+        def _get_master_token() -> str:
+            return self._client.get_master_token() or ""
 
         master_token = await self.hass.async_add_executor_job(_get_master_token)
         if is_aas_et(master_token) is False:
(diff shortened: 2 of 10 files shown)


custom_components/google_home/types.py
```diff
@@ -1,18 +1,5 @@
 """Various types used in type hints."""
-from typing import Iterable, List, Optional, Set, Tuple, TypedDict
-
-from typing_extensions import Protocol
-
-from homeassistant.helpers.entity import Entity
-
-
-class AddEntitiesCallback(Protocol):
```
KapJI, line 9: I realised we don't really need such precise type and simple `Callable` is fine for our case.

DurgNomis-drol:
I have seen that someone is going through home-assistant/core to add typing, so maybe your whish is slowly coming true 😊

KapJI:
I think they were only modernising existing types. There are still hundreds of untyped components 🙁

DurgNomis-drol:
Oh okay, we can only hope that in the future this changes 😄

custom_components/google_home/__init__.py
```diff
@@ -30,7 +30,7 @@
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
 
-async def async_setup(_hass: HomeAssistant, _config: Config) -> bool:
+async def async_setup(_hass: HomeAssistant, _config: dict) -> bool:
```
KapJI, line 33: Btw, this function can be completely removed in 2021.4.0: https://github.com/home-assistant/core/pull/48381

leikoilja: Nice, let's drop it once HA is on the new version.
If we will do so, shall we specify that `ha-google-home` requires HA `2021.4.0` somewhere (perhaps in `hacs.json` file)?

custom_components/google_home/__init__.py
```diff
@@ -76,7 +76,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     # Offload the loading of entities to the platform
     for platform in PLATFORMS:
-        hass.async_add_job(
+        hass.async_create_task(
```
leikoilja, line 79: Why is this change needed? 
from [docs](http://dev-docs.home-assistant.io/en/master/api/core.html):
> 
> async_add_job(target: Callable[[...], Any], *args: Any) → Optional[_asyncio.Future][source]
> Add a job from within the event loop.
> 
> This method must be run in the event loop.
> 
> target: target to call. args: parameters for method to call.
> 
> async_create_task(target: Coroutine) → _asyncio.Task[source]
> Create a task from within the eventloop.
> 
> This method must be run in the event loop.
> 
> target: target to call.

custom_components/google_home/api.py
```diff
@@ -221,9 +221,9 @@ async def update_google_devices_information(self) -> List[GoogleHomeDevice]:
 
         coordinator_data = await gather(
             *[
-                self.get_alarms_and_timers(device, device.ip_address)
+                self.get_alarms_and_timers(device, device.ip_address, device.auth_token)
                 for device in devices
-                if device.ip_address
+                if device.ip_address and device.auth_token
```
leikoilja, line 226: Nice catch

custom_components/google_home/config_flow.py
```diff
@@ -54,8 +53,8 @@ async def async_step_user(self, user_input=None):
                 username=user_input[CONF_USERNAME],
                 password=user_input[CONF_PASSWORD],
             )
-            valid, master_token = await self._test_credentials(client)
-            if valid:
+            master_token = await self._test_credentials(client)
+            if master_token is not None:
```
leikoilja, line 57: Nice catch as well

leikoilja (commented):
Nice! Good work with homeassistant stubs as well! 🚀

custom_components/google_home/__init__.py
KapJI: Yes, that's the plan 🙂

custom_components/google_home/__init__.py
KapJI: It doesn't return `Callable`, I checked other places and they use `async_create_task`: https://github.com/home-assistant/core/blob/f4cc4a0896358f4c8df5a193d900716fabb7ad32/homeassistant/components/xiaomi_miio/__init__.py#L126

leikoilja: Yeh, i guess we had it from [cookiecuter](https://github.com/oncleben31/cookiecutter-homeassistant-custom-component/blob/933ca228d63f5d0c249c9328f16d7f33f6ae1749/%7B%7Bcookiecutter.project_name%7D%7D/custom_components/%7B%7Bcookiecutter.domain_name%7D%7D/__init__.py#L59) where they still use `async_add_job`, but if `async_create_task` is better since it returns `Callable`, as you say, it will probs be good for typing purpose

leikoilja (approved):

PR #153, opened 2021-04-06
Complete strict type checks

### LINKED ISSUE #85: Enable strict typing

KapJI:
- Add missing type hints in `config_flow`. Unfortunately, we still have to use `Any` there.
- Fixed implicit re-export. Now mypy will run this check.
- Found bug with unchecked `None` in `async_delete_alarm`.

Closes #85

### DIFF
--- a/custom_components/google_home/config_flow.py
+++ b/custom_components/google_home/config_flow.py
@@ -34,13 +34,13 @@ class GoogleHomeFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
 
     def __init__(self) -> None:
         """Initialize."""
-        self.errors: Dict[str, str] = {}
+        self._errors: Dict[str, str] = {}
 
     async def async_step_user(
         self, user_input: Optional[Dict[str, Any]] = None
     ) -> Dict[str, Any]:
         """Handle a flow initialized by the user."""
-        self.errors = {}
+        self._errors = {}
 
         # Only a single instance of the integration is allowed:
         if self._async_current_entries():
@@ -65,7 +65,7 @@ class GoogleHomeFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
                     }
                 )
                 return self.async_create_entry(title=username, data=user_input)
-            self.errors["base"] = "auth"
+            self._errors["base"] = "auth"
             return await self._show_config_form(user_input)
 
         return await self._show_config_form(user_input)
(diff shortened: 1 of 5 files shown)


custom_components/google_home/config_flow.py
```diff
@@ -34,11 +34,13 @@ class GoogleHomeFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
 
     def __init__(self) -> None:
         """Initialize."""
-        self._errors: Dict[str, str] = {}
+        self.errors: Dict[str, str] = {}
```
leikoilja, line 37: Why renaming?

custom_components/google_home/config_flow.py
```diff
@@ -106,11 +110,15 @@ def __init__(self, config_entry: ConfigEntry):
         self.config_entry = config_entry
         self.options = dict(config_entry.options)
 
-    async def async_step_init(self, _user_input=None):
+    async def async_step_init(
+        self, _user_input: Optional[Dict[str, Any]] = None
```
leikoilja, line 114: Do you think we'd be able to replace `Dict[str, Any]]` by custom HA stubs that you have been working on?

setup.cfg
```diff
@@ -17,28 +17,15 @@ ignore =
...
-check_untyped_defs = True
-# disallow_any_explicit = True
-# disallow_any_unimported = True
-# disallow_subclassing_any = True
-disallow_incomplete_defs = True
-disallow_untyped_calls = True
-disallow_untyped_decorators = True
-disallow_untyped_defs = True
-no_implicit_optional = True
+strict = True
```
leikoilja, line 20: magical 🤩

leikoilja (approved):
Amazing and hard work! Thanks for working and finishing that epic refactoring 💥 🚀

custom_components/google_home/config_flow.py
KapJI: I was thinking about unused variable convention but this is actually a private field. Python is a bit confusing, I'll revert this.

custom_components/google_home/config_flow.py
KapJI: Not sure, but I can try in another PR.

PR #154, opened 2021-04-06
Update README.md

KapJI:
Simplify table of contents a bit. I don't think we need links to subsubsections.


DurgNomis-drol (approved):
LGTM 😄

PR #155, opened 2021-04-06
Add reboot service

DurgNomis-drol:
This adds a reboot service. 😄 

It is also possible to do a factory reset from the same endpoint but this is not implemented. I can change this if we want this?

I have tested it locally and it works for me 🚀

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -1,10 +1,11 @@
 """Sample API Client."""
 from asyncio import gather
+import json
 import logging
-from typing import List, Optional
+from typing import Any, List, Optional
 
 from aiohttp import ClientError, ClientSession
-from aiohttp.client_exceptions import ClientConnectorError
+from aiohttp.client_exceptions import ClientConnectorError, ContentTypeError
 from glocaltokens.client import Device, GLocalAuthenticationTokens
 from glocaltokens.utils.token import is_aas_et
 from zeroconf import Zeroconf
@@ -18,7 +19,6 @@ from .const import (
     API_ENDPOINT_REBOOT,
     API_RETURNED_UNKNOWN,
     HEADER_CAST_LOCAL_AUTH,
-    HEADER_CONTENT_TYPE,
     HEADERS,
     JSON_ALARM,
     JSON_TIMER,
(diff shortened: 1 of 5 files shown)


custom_components/google_home/sensor.py
```diff
@@ -131,6 +140,16 @@ def get_device_attributes(device: GoogleHomeDevice) -> DeviceAttributes:
             "integration": DOMAIN,
         }
 
+    async def async_reboot_device(self) -> None:
+        """Reboot device"""
```
KapJI, line 144:
```suggestion
        """Reboot the device."""
```

custom_components/google_home/api.py
```diff
@@ -313,7 +314,66 @@ async def delete_alarm_or_timer(
...
+                        response.status,
+                        response,
+                    )
+        except ClientConnectorError:
+            _LOGGER.warning(
+                "Failed to connect to %s device. The device is probably offline.",
+                device.name,
+            )
+        except ClientError as ex:
+            # Make sure that we log the exception from the client if one occurred.
```
KapJI, line 376: These services have a lot of duplicated code. Can it be abstracted with a common function?

Drakulix (approved):
Our api.py really gets a bit repetitive, but that is out of scope for this feature.

Logic seems correct and checks are passing, looks fine to me!

custom_components/google_home/api.py
```diff
@@ -313,7 +314,66 @@ async def delete_alarm_or_timer(
...
         except ClientError as ex:
-            # Make sure that we log the exception if one occurred.
+            # Make sure that we log the exception from the client if one occurred.
+            _LOGGER.error(
+                "Request error: %s",
+                ex,
+            )
+
+    async def reboot_google_device(self, device: GoogleHomeDevice) -> None:
+        """Reboots a Google Home device if supported."""
```
KapJI, line 324:
```suggestion
        """Reboots a Google Home device if it supports this."""
```

custom_components/google_home/api.py
```diff
@@ -313,7 +314,66 @@ async def delete_alarm_or_timer(
...
+
+        if device.ip_address is None:
+            _LOGGER.warning("Device %s doesn't have IP address!", device.name)
+            return
+
+        if device.auth_token is None:
+            _LOGGER.warning("Device %s doesn't have a auth token!", device.name)
```
KapJI, line 331:
```suggestion
            _LOGGER.warning("Device %s doesn't have an auth token!", device.name)
```

KapJI (changes requested):

custom_components/google_home/api.py
DurgNomis-drol: Yes, i thought that too, i even have another thing planned. Should i refactor it in this or should we move it to another PR?

KapJI: I'd rather have it here as this adds a new copy.

KapJI:
If you decide to go with factory reset service, I think it should have a confirmation field, something like `confirmation = "i_want_to_reset"`.

DurgNomis-drol:
@KapJI The reason I haven't added it is exactly because of that scenario. I don't think that it is necessary to add and only mentioned it if some of you had another opinion 😄 But yes, that would be the correct way to do it if so.

KapJI:
I think we can wait before we see someone who needs this for some reason 🙂

custom_components/google_home/api.py
Drakulix: Not sure, if it is really worth it. I want to start replacing api.py with googledevices, once I have finished up another project. While I agree, that some amount de-duplication would likely be beneficial, we should not invest too much here.

KapJI: Well, I think in any case we should try to keep the code here good. I don't think this deduplication should be very hard.

DurgNomis-drol: It is fine, i am nearly done with it. We/you can also use this when making `googledevices`. So I don't see it as a total waste of time. 😄 I am also gonna add a "Do not disturb switch" soon.

DurgNomis-drol:
It is giving me a real headache that we are using strict typing. We can't use `any`, so it is very difficult to make generic function that can handle post requests and return a response that we don't know what contains.

KapJI:
You can enable `Any` for a single line/function if you need it. I can figure out later how to get rid of it.

DurgNomis-drol:
@KapJI how do you do that?

KapJI:
Add `# type: ignore[misc]` on that line.

DurgNomis-drol:
I am having some problems with how to handle the response message from the server. I think i am close to figuring it out, but will push it tomorrow 😪

KapJI:
You can push unfinished version, maybe I can help.

DurgNomis-drol:
I actually just got it working, had to check mime_type. But it should work now. 😄 

It is logging `NotImplementedError` for `remove_service` every time you call `reboot_device`, is there a fix in the works over at `glocaltokens`?

I removed `charset=UTF8` from headers as i don't think we need it, just a heads up.

KapJI:
`remove_service` was added in https://github.com/leikoilja/glocaltokens/pull/53, we just need to update the dependency.

@leikoilja let's merge https://github.com/leikoilja/glocaltokens/pull/63 and publish the new version.

custom_components/google_home/api.py
```diff
@@ -236,73 +237,108 @@ async def delete_alarm_or_timer(
...
+        if response:
+            # It will return true even if the device does not support rebooting.
+            _LOGGER.info(
+                "Successfully asked %s to reboot.",
+                device.name,
+            )
+
+    async def post(  # type: ignore[misc]
+        self, endpoint: str, data: str, device: GoogleHomeDevice
+    ) -> Any:
```
KapJI, line 306: I think this should work. Should be enough precision and better that `Any`.
```suggestion
    ) -> Optional[Dict[str, str]]:
```

custom_components/google_home/const.py
```diff
@@ -54,13 +55,14 @@
...
 API_ENDPOINT_DELETE = "setup/assistant/alarms/delete"
+API_ENDPOINT_REBOOT = "setup/reboot"
 HEADER_CAST_LOCAL_AUTH = "cast-local-authorization-token"
 HEADER_CONTENT_TYPE = "content-type"
 
 # HEADERS
 HEADERS = {
     HEADER_CAST_LOCAL_AUTH: "",
-    HEADER_CONTENT_TYPE: "application/json; charset=UTF-8",
+    HEADER_CONTENT_TYPE: "application/json",
```
KapJI, line 65: Probably this is default but I remember you said you can assign name to an alarm? Can you check if UTF-8 name works fine without setting `charset`?

custom_components/google_home/api.py
DurgNomis-drol: I will test it 😊

DurgNomis-drol: Worked! Thanks for the help

custom_components/google_home/const.py
DurgNomis-drol: It works, i even tried with the special letters we have here in Denmark and it works just fine. It was maybe me who added it sometime ago with out fully understanding that it was not needed. 😄 

[`Ghlocalapi`](https://rithvikvibhu.github.io/GHLocalApi/#section/Google-Home-Local-API/Getting-Started) also does not say anything about it 😄

custom_components/google_home/api.py
```diff
@@ -236,73 +237,108 @@ async def delete_alarm_or_timer(
...
             url,
-            data,
         )
 
+        resp = None
+
         try:
             async with self._session.post(
-                url, json=data, headers=HEADERS, timeout=TIMEOUT
+                url, data=data, headers=HEADERS, timeout=TIMEOUT
```
KapJI, line 332: Why did you change `json` to `data`? Is it also works?

KapJI: Documentation says that it only adds "content-type: application/json", but I think setting headers explicitly redefines this anyway so it should be fine.

KapJI (approved):
Looks great, thanks! 🎉

custom_components/google_home/api.py
DurgNomis-drol: `json` also converts it to JSON as far as i understand.

It is because of strict typing. If we use `json` we need to pass a known dict and we don't know this. This way we only need to define it to a `str`.

But maybe i am wrong about this. 🤔

leikoilja:
Awesome job guys, @DurgNomis-drol, @KapJI! Thanks for implementing it 💥 
I'll make a new release of `glocaltokens` today and open a PR here to bump it up as well ;)

DurgNomis-drol:
@leikoilja Awesome! 💥

custom_components/google_home/api.py
KapJI: I see. Yeah, that's not easy to construct proper type for json and mypy still doesn't support recursive types which are needed for the correct type.

DurgNomis-drol: This is a common problem i am running into with different libraries unfortunately. 😢 But by finding a fix I only get better at understanding python. 🚀 Also many places they use `Any` instead of explicitly stating the correct type.

PR #158, opened 2021-04-07
Improve typing in config flow

KapJI:
I was looking at removing `Dict[str, Any]` in config flow and this is as far as we can go now. Other places require changing HA API, I'll try to do this at some point.

But also I found some small issues here:
- `data = user_input` created new reference to the same dict and it was updated while the original reference was passed to the `async_create_entry`. It was working as expected but I don't get why we need to do that is such way.

### DIFF
--- a/custom_components/google_home/config_flow.py
+++ b/custom_components/google_home/config_flow.py
@@ -22,6 +22,7 @@ from .const import (
     DOMAIN,
 )
 from .exceptions import InvalidMasterToken
+from .types import OptionsFlowDict
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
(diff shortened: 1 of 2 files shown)


KapJI:
@leikoilja do you have any concerns about this?

custom_components/google_home/config_flow.py
```diff
@@ -57,17 +58,10 @@ async def async_step_user(
             )
             master_token = await self._test_credentials(client)
             if master_token is not None:
-                data = user_input
```
leikoilja, line 60: It is indeed a weird thing that we are doing here 🤔

leikoilja (approved):
Looks great, thanks 🚀

PR #166, opened 2021-04-09
catch all exceptions in api.py Fetching data loop.

vlebourl:
raise when it's a dangerous exception.

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -197,6 +197,10 @@ class GlocaltokensApiClient:
                 ex,
             )
             device.available = False
+        except Exception as e:
+            import sys
+            _LOGGER.error(" /!\ ERROR: %s", sys.exc_info()[0])
+            raise
         return device
 
     async def update_google_devices_information(self) -> List[GoogleHomeDevice]:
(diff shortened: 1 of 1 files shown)


KapJI:
What exception did you get which is not yet catched?

vlebourl:
sorry... going to fast with my commits...
I get this:
```
/!\ ERROR: <class 'asyncio.exceptions.TimeoutError'>
```
which I know, is printed by HA like this:
```
2021-04-09 17:40:20 ERROR (MainThread) [custom_components.google_home] Timeout fetching sensor data
```
later on, but I thought that way we can know where the timeout precisely comes from.
If you judge that unnecessary, please do not hesitate to discard that PR. I'll keep that change private ;-)

vlebourl:
Also, one of my device is constantly in a timeout, maybe allowing to set the timeout as an option in the config flow could be helpful. but I guess that would be another PR!

DurgNomis-drol:
@vlebourl First of all you have to use pre-commit and make sure that everything passes if you want your PR to be reviewed. 😄 See  [Contributions.md](https://github.com/leikoilja/ha-google-home/blob/master/CONTRIBUTING.mdl.

Second, this just print there error. We still need a way to handle it. But first we need to understand why it is happening with some device and not others.

KapJI:
We shouldn't actually catch generic `Exception`. Instead we need to list exceptions which we're expecting to get. Everything else should go to the top level and be reported as a bug.

PR #172, opened 2021-04-10
Fix do not disturb switch

DurgNomis-drol:
Small PR to fix switch not working.

I also moved JSON attribute to a CONST.

@leikoilja Any thing else from the other PR review? 🤔 

@KapJI About it blinking I think i know how to fix that. but will make a separate PR for that. 😄

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -332,16 +332,16 @@ class GlocaltokensApiClient:
 
         data = None
 
-        if enable is None:
-            _LOGGER.debug(
-                "Getting Do Not Disturb setting from Google Home device %s",
-                device.name,
-            )
-        else:
+        if enable is not None:
             data = {"notifications_enabled": not enable}
             _LOGGER.debug(
                 "Setting Do Not Disturb setting to %s on Google Home device %s",
-                not enable,
+                enable,
+                device.name,
+            )
+        else:
+            _LOGGER.debug(
+                "Getting Do Not Disturb setting from Google Home device %s",
                 device.name,
             )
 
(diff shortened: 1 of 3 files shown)


KapJI:
There was also a README change 🙂

custom_components/google_home/const.py
```diff
@@ -84,6 +84,7 @@
 # JSON parameter values when retrieving information from devices
 JSON_ALARM = "alarm"
 JSON_TIMER = "timer"
+JSON_NOTIFICATION_ENABLED = "notifications_enabled"
```
KapJI, line 87:
```suggestion
JSON_NOTIFICATIONS_ENABLED = "notifications_enabled"
```

KapJI (approved):

DurgNomis-drol:
Thanks @KapJI for approving it 😃

custom_components/google_home/api.py
```diff
@@ -332,33 +333,34 @@ async def update_do_not_disturb(
 
         data = None
 
-        if enable is None:
+        if enable is not None:
```
leikoilja, line 336: This could just be `if enable`, right? :)

DurgNomis-drol: I tried that, but if the value is false, python just moves to else instead. 😊

leikoilja: hm... but do we expect `True` here? Maybe then `if enable is True`? 😅

DurgNomis-drol: We expected True or False if we want to set the status on the device or None if we just want get status from device 😊

leikoilja: ouuu gotcha now. Thanks for explaining 👍

DurgNomis-drol: No problem 😊

PR #175, opened 2021-04-11
Stricter Labeler rules

KapJI:
Right now Labeler is more annoying than useful. Reduce number of false positives by adding labels only to PRs where no python code is changed.


leikoilja (approved):
Yeh, unfortunately you're right, somehow it introduced tons of false positives, let's hope this will help it :)

PR #180, opened 2021-04-11
Fix labeler config

KapJI:
It seems if 2 rules match, none is aplied.


leikoilja (approved):
haha, i'm starting to doubt the value of labeler at all :D

PR #189, opened 2021-04-12
Fix Labeler

KapJI:
One more time. I think now I understand it better. Need to apply rules recursively.


leikoilja (approved):
Hahha, we need ppl like you to write getter README documentation and examples on https://github.com/actions/labeler

PR #205, opened 2021-04-14
Put inactive alarms to the end of sorted list

### LINKED ISSUE #204: Display Home devices keep old alarms thus "next alarm" is always the same old alarm in the sensor state
## Describe the bug
The Google home devices with screen at least save all old alarms even after dismissed. Because of this the first alarm in the sensor state is the first one I ever set and can be weeks/months old and of no interest. It is only future alarms that are of interest in most usecases. The sensor should show the next upcoming alarm instead.

## Version of the integration 1.4.2

```text





```

## Logs

```text
Add your logs here.
```

KapJI:
- Add them to the end but still sorted.
- If next alarm is inactive, set state to `unavailable`.

Fixes #204

### DIFF
--- a/custom_components/google_home/models.py
+++ b/custom_components/google_home/models.py
@@ -69,8 +69,13 @@ class GoogleHomeDevice:
         ]
 
     def get_sorted_alarms(self) -> List[GoogleHomeAlarm]:
-        """Returns alarms in a sorted order"""
-        return sorted(self._alarms, key=lambda k: k.fire_time)
+        """Returns alarms in a sorted order. Inactive alarms are in the end."""
+        return sorted(
+            self._alarms,
+            key=lambda k: k.fire_time
+            if k.status != GoogleHomeAlarmStatus.INACTIVE
+            else k.fire_time + sys.maxsize,
+        )
 
     def get_next_alarm(self) -> Optional[GoogleHomeAlarm]:
         """Returns next alarm"""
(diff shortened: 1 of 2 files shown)


custom_components/google_home/models.py
```diff
@@ -69,8 +69,13 @@ def set_timers(self, timers: List[TimerJsonDict]) -> None:
...
     def get_sorted_alarms(self) -> List[GoogleHomeAlarm]:
-        """Returns alarms in a sorted order"""
-        return sorted(self._alarms, key=lambda k: k.fire_time)
+        """Returns alarms in a sorted order. Inactive alarms are in the end."""
+        return sorted(
+            self._alarms,
+            key=lambda k: k.fire_time
+            if k.status != GoogleHomeAlarmStatus.INACTIVE
+            else k.fire_time + sys.maxsize,
+        )
```
leikoilja, line 78: oh i really want to start introducing smaller unit tests for features like that to make sure they work 😅

leikoilja (approved):
Hard to tell if this will work fine, but it looks good 👍  Thanks for fixing it.

custom_components/google_home/models.py
KapJI: you're welcome to add them 😀

leikoilja: one day i ll get my hands on them :D

PR #209, opened 2021-04-15
Setup codespell

KapJI:
Codespell can find typos without being annoying. HA also uses it.


leikoilja (approved):
Awesome find! Thanks for implementing it, we need that 😅

Issue #211, opened 2021-04-15
Stop fetching or alarms and timers when device is "Chromecast *"


stboch:
## Describe the bug

Since this isn't currently a supported feature in chromecast devices, and the device hardware is known by the app, perhaps rather then retrying at next interval only do it when the component reloads/starts

## Version of the integration

```text
1.4.2
```

## Logs

```text
2021-04-15 13:28:55 DEBUG (MainThread) [custom_components.google_home] Failed to post data to Master Bedroom TV 2, API returned 404. The device(hardware='Chromecast') is possibly not Google Home compatible and has no alarms/timers. Will retry later.
2021-04-15 13:28:55 DEBUG (MainThread) [custom_components.google_home] Failed to fetch data from Living Room TV 2, API returned 404. The device(hardware='Chromecast Ultra') is possibly not Google Home compatible and has no alarms/timers. Will retry later.
```

stboch:
similar to #118

PR #212, opened 2021-04-15
Add exclusions for Chromecast devices

stboch:
Since chromecast devices do not currently support the google home api, exclude these devices from api calls in the same manner as devices missing IP's or auth tokens.

PR to address issue #211

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -17,6 +17,7 @@ from .const import (
     API_ENDPOINT_DELETE,
     API_ENDPOINT_DO_NOT_DISTURB,
     API_ENDPOINT_REBOOT,
+    HARDWARE_CHROMECAST,
     HEADER_CAST_LOCAL_AUTH,
     HEADER_CONTENT_TYPE,
     JSON_ALARM,
@@ -24,7 +25,6 @@ from .const import (
     JSON_TIMER,
     PORT,
     TIMEOUT,
-    HARDWARE_CHROMECAST,
 )
 from .exceptions import InvalidMasterToken
 from .models import GoogleHomeDevice
@@ -131,8 +131,8 @@ class GlocaltokensApiClient:
                     ),
                     device.name,
                 )
-            if device.hardware in HARDWARE_CHROMECAST :
-                device.available = False 
+            if device.hardware in HARDWARE_CHROMECAST:
+                device.available = False
                 _LOGGER.debug(
                     (
                         "The device %s (hardware='%s') is not Google Home "
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -24,6 +24,7 @@ from .const import (
     JSON_TIMER,
     PORT,
     TIMEOUT,
+    HARDWARE_CHROMECAST,
 )
 from .exceptions import InvalidMasterToken
 from .models import GoogleHomeDevice
@@ -130,6 +131,16 @@ class GlocaltokensApiClient:
                     ),
                     device.name,
                 )
+            if device.hardware in HARDWARE_CHROMECAST :
+                device.available = False 
+                _LOGGER.debug(
+                    (
+                        "The device %s (hardware='%s') is not Google Home "
+                        "compatible and is currently unsupported. "
+                    ),
+                    device.name,
+                    device.hardware,
+                )
 
         coordinator_data = await asyncio.gather(
             *[
@@ -137,7 +148,7 @@ class GlocaltokensApiClient:
                     device=device,
                 )
                 for device in devices
-                if device.ip_address and device.auth_token
+                if device.ip_address and device.auth_token and device.available
             ]
         )
         return coordinator_data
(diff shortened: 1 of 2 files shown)


KapJI (changes requested):
Chromecast devices do support Google Home local API. They don't support alarms. We still want to create `device` sensor for them.

The idea is that we don't want to have explicit list of supported devices. If device responds to alarms endpoint, it's supported otherwise it's not. Currently we keep polling unsupported devices and this can be improved. Like stop polling devices which returned 404.

I'm not sure what's your concern here. These log messages you have are on debug level anyway. I'm not convinced we need this.

stboch:
Really just trying to cut back on all the erroneous calls to stuff that will fail i have about 12 chromecasts on my network just a lot of traffic that will fail. 
Either way my rush realized that using the other method I moved to higher in the calls wasn't the best. would be better implemented in the collect_data_from_endpoints as it will still populate the other components in HA just lack the other entities.

KapJI:
My calculations that each request is under 1 KB. With polling interval of 10 s, you have ~1 KB/s of erroneous traffic in your network. Shouldn't be a big deal.

leikoilja:
@KapJI i still think we need to come up with a better way to sort of entities as users ask. As i mentioned somewhere else, we can try doing requests to `eureka` endpoint and fetching device capabilities and based on that filter the ones that don't support google assistant. I tried playing with that a bit the other day and think it could be doable

KapJI:
@leikoilja even without eureka info chromecasts return 404 on alarms endpoint. We can use that to filter them out without creating an explicit model list.

leikoilja:
Either that or reverting back to the config options where one can add custom filtering wildcard(s)

DurgNomis-drol:
I agree with @leikoilja and @stboch I have an android TV and it constantly makes request to it for alarms and timers, and Do Not Disturb which is not supported and we should find a way to eliminated unnecessary request, even if they are small and fast. 

As far as i understand, we can get determine if the devices is compatible through the eureka info endpoint.

KapJI:
I think just adding device to unsopported list if it returns 404 is the simplest and the most robust way to do that. 🙂

DurgNomis-drol:
But what if a device is offline? Then it will return 404, but still be compatible.

KapJI:
Offline revice can't return 404. It will be unreachable.

DurgNomis-drol:
Right, i think i understand now.

Or we can just check it against this information returned from `eureka_info` endpoint:

```yaml
"device_info": {
  "capabilities": {
    "aogh_supported": true,
    "assistant_supported": true,
    "audio_hdr_supported": false,
    "audio_surround_mode_supported": false,
    "ble_supported": true,
    "bluetooth_audio_sink_supported": true,
    "bluetooth_audio_source_supported": true,
    "bluetooth_supported": true,
    "cloudcast_supported": true,
    "content_filters_supported": true,
    "display_supported": false,
    "fdr_supported": false,
    "hdmi_prefer_50hz_supported": false,
    "hdmi_prefer_high_fps_supported": false,
    "hotspot_supported": true,
    "https_setup_supported": true,
    "input_management_supported": true,
    "keep_hotspot_until_connected_supported": true,
    "multi_user_supported": true,
    "multichannel_group_supported": true,
    "multizone_supported": true,
    "night_mode_supported": true,
    "night_mode_supported_v2": true,
    "opencast_supported": false,
    "preview_channel_supported": true,
    "reboot_supported": true,
    "remote_ducking_supported": true,
    "separate_tts_volume_supported": true,
    "setup_supported": true,
    "sleep_mode_supported": true,
    "stats_supported": true,
    "system_sound_effects_supported": false,
    "user_eq_supported": true,
    "wifi_auto_save_supported": true,
    "wifi_regulatory_domain_locked": true,
    "wifi_supported": true
    },
}
```

I believe that `assistant_supported` is the one we are looking for. 😃

KapJI:
Yeah, checking for 404 or querying `eureka_info`, both should work. My idea is checking 404 is simpler and more straightforward as you don't need to query another endpoint.

DurgNomis-drol:
Actually even if the devices returns a `200` for `eureka_info`, there is no guaranty that the device supports alarms or timers, we still would need to check `assistant_supported`. This can also come in handy in the future where we can check what other features is supported. 

The check should probably be a combination of both like this:

1. Request `eureka_info` if `404` return device **not** compatible. Else continue check.
2. Check if `assistant_supported` is true, return device is compatible. If false return device is **not** compatible.

KapJI:
I mean 404 on `alarms` endpoint, not `eureka_info`.

DurgNomis-drol:
Mine is more explicit, which is very pythonic. 😄

stboch:
using the array of supported features is a better result and is what we do on the surbaru side, we map the returned feature codes a car has to the APIs that we support to prevent this calls. Now that's to prevent calls to a third party site and not get caught being naughty to the api but the same effect locally is still a good choice. If all local devices still respond to eureka_info and we can differentiate the APIs based on the supported status then I think this is the right way to go.

DurgNomis-drol:
I have created a separate issue to track that feature. 😃 Lets keep all further discussions about over there.

PR #218, opened 2021-04-16
Modernise type hints

KapJI:
[python-typing-update](https://github.com/cdce8p/python-typing-update) is created by HA maintainers to update type hints to the new [PEP 585](https://www.python.org/dev/peps/pep-0585/) style.

Old ones will be deprecated in Python 3.9 and eventually removed few year later. Before that they can be used with:
```python
from __future__ import annotations
```

The advantage is that they don't need to be imported and python authors don't need to maintain parallel type definitions in `typing` module. Basically this is the future of python.

This tool can be run with:
```shell
poetry run pre-commit run --hook-stage manual python-typing-update --all-files
```

It can only run on committed files, because of that we can't add it to default `pre-commit` workflow. Otherwise you want be able to commit files if it modified something. But can add it to linting job.

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -33,3 +33,7 @@ jobs:
       - name: Run pre-commit on all files
         run: |
           poetry run pre-commit run --all-files --show-diff-on-failure --color=always
+
+      - name: Run python-typing-update
+        run: |
+          poetry run pre-commit run --hook-stage manual python-typing-update --all-files --show-diff-on-failure --color=always
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,3 +45,16 @@ repos:
     rev: v2.2.1
     hooks:
       - id: prettier
+  - repo: https://github.com/cdce8p/python-typing-update
+    rev: v0.3.3
+    hooks:
+      # Run `python-typing-update` hook manually from time to time
+      # to update python typing syntax.
+      # Will require manual work, before submitting changes!
+      - id: python-typing-update
+        stages: [manual]
+        args:
+          - --py38-plus
+          - --force
+          - --keep-updates
+        files: ^.+\.py$
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -1,7 +1,9 @@
 """Sample API Client."""
+from __future__ import annotations
+
 import asyncio
 import logging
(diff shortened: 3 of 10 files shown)


DurgNomis-drol:
A couple of questions:
1. Will this require us to run `poetry run pre-commit run --hook-stage manual python-typing-update --all-files` before every commit?
2. Is this just another syntax for using typing?

KapJI:
1. You don't have to as long you don't forget to use these modern syntax. Otherwise CI will catch it.
2. Yes, it's added by [PEP 585](https://www.python.org/dev/peps/pep-0585/).

DurgNomis-drol:
Right, thanks clarifying. 

It seems cleaner to use the new type hints then 😃

KapJI:
The only limitation of this new syntax is that it can only be used for type hints but can't be used in the code itself (e.g. for `isinstance`, `cast` or type aliases like `JsonDict`) before Python 3.9 is the default. HA is planning to switch to 3.9 in the end of this year.

DurgNomis-drol:
Right, so we will stil need to use the old syntax in the code? 

If above is the case, then I think it will become confusing for people if we use two different syntax's for typing.

KapJI:
In very limited cases. I think right now `JsonDict` is the only place where it's needed. This is the style I'm trying to adopt from HA Core codebase.

DurgNomis-drol:
HA core does not use type annotations in the code itself. There by not having the same problem as us. 

What does @leikoilja think?

KapJI:
> HA core does not use type annotations in the code itself.
That's not true. Open any file here for example.
https://github.com/home-assistant/core/tree/dev/homeassistant/helpers

DurgNomis-drol:
I think I misunderstood what you meant with `can't be used in the code itself`.

KapJI:
I mean code which is executed by python interpreter.

For example type hints are not executed, only `mypy` uses them.
But on the other hand `isinstance(List[str], my_list)` is code which is executed by interpreter.

DurgNomis-drol:
Thanks for clarifying, I understand it now. 😄 

This sounds great!

leikoilja (approved):
This is very nice! 
Typing can be a bit confusing at time, and i'm also just grasping my idea around it, so as long as we have someone like @KapJI who can take a solid lead on it - I'm all hands for it! 🚀

PR #235, opened 2021-04-29
Set IoT class

KapJI:
Custom components are now required to have it set.


leikoilja (approved):
Awesome! Thanks for adding it 🚀

leikoilja (approved):

PR #250, opened 2021-06-06
Alarm volume

ArnyminerZ:
Added an alarm volume entity

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -16,7 +16,7 @@ from homeassistant.core import HomeAssistant
 
 from .const import (
     API_ENDPOINT_ALARMS,
-    API_ENDPOINT_DELETE,
+    API_ENDPOINT_ALARM_DELETE,
     API_ENDPOINT_DO_NOT_DISTURB,
     API_ENDPOINT_REBOOT,
     HEADER_CAST_LOCAL_AUTH,
@@ -25,7 +25,7 @@ from .const import (
     JSON_NOTIFICATIONS_ENABLED,
     JSON_TIMER,
     PORT,
-    TIMEOUT,
+    TIMEOUT, JSON_ALARM_VOLUME, API_ENDPOINT_ALARM_VOLUME,
 )
 from .exceptions import InvalidMasterToken
 from .models import GoogleHomeDevice
@@ -195,7 +195,7 @@ class GlocaltokensApiClient:
         )
 
         response = await self.request(
-            method="POST", endpoint=API_ENDPOINT_DELETE, device=device, data=data
+            method="POST", endpoint=API_ENDPOINT_ALARM_DELETE, device=device, data=data
         )
 
         if response is not None:
@@ -248,7 +248,7 @@ class GlocaltokensApiClient:
             )
 
     async def update_do_not_disturb(
-        self, device: GoogleHomeDevice, enable: bool | None = None
+            self, device: GoogleHomeDevice, enable: bool | None = None
     ) -> GoogleHomeDevice:
         """Gets or sets the do not disturb setting on a Google Home device."""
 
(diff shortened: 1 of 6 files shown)


ArnyminerZ:
Still doing some tests to check if everything is set up correctly

custom_components/google_home/api.py
```diff
@@ -297,6 +299,57 @@ async def update_do_not_disturb(
...
+            method="POST",
+            endpoint=API_ENDPOINT_ALARM_VOLUME,
+            device=device,
+            data=data,
+            polling=polling,
+        )
+        if response is not None:
+            if JSON_NOTIFICATIONS_ENABLED in response:
+                volume_raw: str = str(response[JSON_ALARM_VOLUME])
+                loaded_volume: float = float(volume_raw)
```
KapJI, line 335: No need to repeat that, it'll be inferred.
```suggestion
                volume_raw = str(response[JSON_ALARM_VOLUME])
                loaded_volume = float(volume_raw)
```

custom_components/google_home/const.py
```diff
@@ -83,6 +89,10 @@
 JSON_ALARM: Final[str] = "alarm"
 JSON_TIMER: Final[str] = "timer"
 JSON_NOTIFICATIONS_ENABLED: Final[str] = "notifications_enabled"
+JSON_ALARM_VOLUME: Final[str] = "volume"
+
+# Device class names
+DEVICE_CLASS_ALARM_VOLUME = "google_home__alarm_volume"
```
KapJI, line 95: This is not a valid device class.

custom_components/google_home/models.py
```diff
@@ -193,3 +203,6 @@ class GoogleHomeTimerStatus(Enum):
     SET = 1
     PAUSED = 2
     RINGING = 3
+
+
+GOOGLE_HOME_ALARM_DEFAULT_VALUE: Final[float] = 1
```
KapJI, line 208: Move to `const.py`.
```suggestion
GOOGLE_HOME_ALARM_DEFAULT_VALUE: Final[float] = 0
```

custom_components/google_home/number.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+async def async_setup_entry(
+    hass: HomeAssistant,
+    entry: ConfigEntry,
+    async_add_devices: Callable[[Iterable[NumberEntity]], None],
+) -> bool:
+    """Setup switch platform."""
+    client = hass.data[DOMAIN][entry.entry_id][DATA_CLIENT]
+    coordinator = hass.data[DOMAIN][entry.entry_id][DATA_COORDINATOR]
```
KapJI, line 37: Here type hints can be useful.
```suggestion
    client: GlocaltokensApiClient = hass.data[DOMAIN][entry.entry_id][DATA_CLIENT]
    coordinator: DataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id][DATA_COORDINATOR]
```

custom_components/google_home/number.py
```diff
@@ -0,0 +1,125 @@
...
+    def unit_of_measurement(self) -> str | None:
+        """The unit of measurement for the entity"""
+        return PERCENTAGE
+
+    @property
+    def icon(self) -> str:
+        """Return the icon of the sensor."""
+        device = self.get_device()
+        if device is None:
+            return ICON_ALARM_VOLUME_HIGH
```
KapJI, line 74: I think this is a better default.
```suggestion
            return ICON_ALARM_VOLUME_OFF
```

custom_components/google_home/number.py
```diff
@@ -0,0 +1,125 @@
...
+        device = self.get_device()
+
+        if device is None:
+            return GOOGLE_HOME_ALARM_DEFAULT_VALUE
+
+        volume = device.get_alarm_volume()
+        return volume
+
+    @property
+    def device_state_attributes(self) -> AlarmVolumeAttributes:
```
KapJI, line 111:
```suggestion
    def extra_state_attributes(self) -> AlarmVolumeAttributes:
```

custom_components/google_home/number.py
```diff
@@ -0,0 +1,125 @@
...
+        """Return the state attributes."""
+        return {
+            "device_class": DEVICE_CLASS_ALARM_VOLUME,
+            "integration": DOMAIN,
```
KapJI, line 115: Is it required?

custom_components/google_home/api.py
```diff
@@ -297,6 +299,57 @@ async def update_do_not_disturb(
...
+            )
+
+        response = await self.request(
+            method="POST",
+            endpoint=API_ENDPOINT_ALARM_VOLUME,
+            device=device,
+            data=data,
+            polling=polling,
+        )
+        if response is not None:
```
leikoilja, line 332:
```suggestion
        if response:
```

leikoilja (commented):
Awesome stuff, thank you @ArnyminerZ 🚀

custom_components/google_home/number.py
ArnyminerZ: Well, not required, but recommended for localization, as far as my investigation has gone. Even though translation may not be that necessary in this case.

KapJI:
There is some pylint error. Probably just disable `duplicate-code` check in `pyproject.toml`.

Also I suggest using rebase instead of merge.

KapJI:
This needs to be documented in README.

custom_components/google_home/api.py
```diff
@@ -149,6 +151,7 @@ async def collect_data_from_endpoints(
     ) -> GoogleHomeDevice:
         """Collect data from different endpoints."""
         device = await self.update_alarms_and_timers(device)
+        device = await self.update_alarm_volume(device)
```
KapJI, line 154: Yes, this is required to keep the value in sync if it's changed from the Google Home app or by voice command.

KapJI (approved):
Awesome, thanks! 🎉

KapJI:
There is a merge conflict which needs to be resolved.

KapJI:
The indices are shifted a bit 🙂

ArnyminerZ:
OMG this check is driving me crazy, so picky 😂

leikoilja:
@ArnyminerZ, are you using `pre-commit` hooks locally (they run the same checks, but locally on your computer)? They can really help to quickly scan the code even before committing it :)

ArnyminerZ:
> @ArnyminerZ, are you using `pre-commit` hooks locally (they run the same checks, but locally on your computer)?

Yes I do, but my WSL machine is giving some problems, so some of them had to be manually fixed.

Issue #233, opened 2021-04-28
Input-text overflows device_card in configuration settings


Mariusthvdb:
## Describe the bug

<img width="1084" alt="Schermafbeelding 2021-04-28 om 14 56 02" src="https://user-images.githubusercontent.com/33354141/116407484-1a6b4180-a832-11eb-807d-b01a5b0d1804.png">

## Version of the integration

If you are unsure about the version check the [github release here](https://github.com/leikoilja/ha-google-home/releases).

```text
1.4.3
```

## Logs

```text
Add your logs here.
```
Nothing in the logs, just an huge overflow on all devices the integration has succesfully discovered.

KapJI:
Yeah, this is known issue. We decided to change state of device to ip address, need someone to implement this.

ZTHawk:
Hello,

I have the same issue. Setting "name" does not help either.

```
- entity: sensor.nest_mini_device
  name: mydevice
- entity: sensor.nest_mini_alarms
  name: myalarm
- entity: sensor.nest_mini_timers
  name: mytimer
- entity: switch.nest_mini_do_not_disturb
  name: mydisturb
```

Everything except "mydevice" works. That way I have an entity with a weird name in Lovelace which cannot be changed.

PR #251, opened 2021-06-06
Changed device status

ArnyminerZ:
Changed the device status from the key to the IP address as stated in #233 because of overflowing.


KapJI:
Please update README as well.

KapJI (approved):
Thanks!

PR #265, opened 2021-06-12
Adding translation label and release drafter

leikoilja:
- Adding translation label
- Adding translation category to release drafter
- Adding translation labeler


.github/labeler.yml
```diff
@@ -9,3 +9,7 @@ documentation:
 github structure:
   - any: [".github/**/*"]
     all: ["!**/*.py"]
+
+translation:
+  - any: ["custom_components/google_home/translations/*"]
+    all: ["!**/*.py"]
```
ArnyminerZ, line 15: Why adding this? Translations will always be in .json, don't they?

leikoilja: Yes they would, but I think it's better to set the explicit path to the `json` files so if we would to introduce any more `json` files and we change them, the GH won't mark them as "translation" label. For example, we have a file `hacs.json` but when that's modified we don't want to mark a PR with "translation" label :)

ArnyminerZ: I meant the `all` tag.

KapJI, line 15: I agree with @ArnyminerZ, this should be cleaner
```suggestion
  - any: ["custom_components/google_home/translations/*.json"]
```

ArnyminerZ (approved):

leikoilja:
Awesome, thanks boys 🤩💥🔥

PR #266, opened 2021-06-12
[WIP] Added eureka fetching

ArnyminerZ:
Added Eureka data fetching. Right now only `device_info` gets loaded, which should be enough for getting `mac_address` to fix the multiple ID issue, and check compatibility.


ArnyminerZ:
Quite a dirty code, maybe you find a different way of loading everything

custom_components/google_home/const.py
```diff
@@ -89,8 +90,128 @@
...
+# EUREKA
+EUREKA_KEY_PARAMS = "params"
+EUREKA_KEY_OPTIONS = "options"
+EUREKA_PARAMS = (
+    "version,audio,name,build_info,detail,device_info,net,wifi,setup,settings,"
+    "opt_in,opencast,multizone,proxy,night_mode_params,user_eq,room_equalizer,"
+    "sign,aogh,ultrasound,mesh "
+)
+EUREKA_OPTIONS = "detail"
+EUREKA_JSON_DEVICE_INFO_4K_BLOCKED = "4k_blocked"
```
KapJI, line 107: I think it's cleaner to define non final TypedDict for returned JSON instead all of these. It can be also used instead of `EurekaCompatibilityDict`. Should allow to remove significant part of the code.

custom_components/google_home/const.py
```diff
@@ -89,8 +90,128 @@
...
 JSON_ALARM: Final[str] = "alarm"
 JSON_TIMER: Final[str] = "timer"
-JSON_NOTIFICATIONS_ENABLED: Final[str] = "notifications_enabled"
 JSON_ALARM_VOLUME: Final[str] = "volume"
+JSON_NOTIFICATIONS_ENABLED: Final[str] = "notifications_enabled"
+JSON_EUREKA_DEVICE_INFO: Final[str] = "device_info"
+JSON_EUREKA_CAPABILITIES: Final[str] = "capabilities"
+
+# EUREKA
+EUREKA_KEY_PARAMS = "params"
```
KapJI, line 99: Actually just `Final` should be enough, you may strip all `[str]` here.
```suggestion
EUREKA_KEY_PARAMS: Final[str] = "params"
```

custom_components/google_home/models.py
```diff
@@ -185,6 +212,327 @@ def as_dict(self) -> GoogleHomeAlarmDict:
...
+        self.device_info = device_info
+
+    def as_dict(self) -> GoogleHomeEurekaDict:
+        """Return typed dict representation."""
+        return {
+            "device_info": self.device_info.as_dict(),
+        }
+
+
+class EurekaDeviceInfo:
```
KapJI, line 228: What's the point of this class compare to just using original TypedDict?

KapJI (changes requested):

Leatherface75:
I see this code is able to get MAC address for devices.
It would be nice if it could get rssi information for devices too.

leikoilja:
Let's close this stale PR for now, but feel free to re-open if someone continues the work on implementing eurika fetching :)

PR #267, opened 2021-06-13
Update Russian translation

KapJI:
- Fix grammar and punctuation, improve wording.
- Use terms from Google docs.
- Remove "Restart requirement" from all translations because it's not implemented and we should be able to implement it without restart.


leikoilja (approved):
Awesome that's exactly what I was talking about in the initial Russian translation PR. Thanks 🤩💥

PR #279, opened 2021-06-26
Changed default status from unavailable to none.

### LINKED ISSUE #278: Timer / Alarm entities set to Unavailable if no timers/alarms, should be none
When no timer is set the status of the *_timers entity is currently set to "unavailable". The ame is true for the *_alarms entity when no alarms. 

Unavailable is considered an error state in HA, so the entities get flagged (and in my case show up in my issues listing) 

Please change the status to "none" instead, and onl to unavailable it the device can not be reached

ArnyminerZ:
Should fix #278. Not sure about the translations file, but as far as I have found in the [Home Assistant Docs](https://developers.home-assistant.io/docs/internationalization/core/) it should be done this way.

Take a look and see if it's correct 😄


KapJI (changes requested):
No, it's not correct.

The reason why it's `unavailable` is that it's compatible with timestamp device class. It only support `unknown` and `unavailable` as valid values which are not a timestamp. You can see that `none` is rendered in the UI incorrectly, something like "invalid format".

I believe that HA's android alarm sensor behaves the same way.

Issue #202, opened 2021-04-13
Device timesout after a getting data for a short time


DurgNomis-drol:
## Describe the bug

Hey guys.

I have for the last week or so been having a weird problem with my Google Nest Mini. Our integration works great, but after a couple of minutes i get this debug message (See logs below) which i expect.

I can still talk to the device, but in the Google Home app i cannot access my device settings for my device. My first thought was that something had gone wrong with the device. I tried reseting it multiple times. When i disable or stop our integration it seems to work again, but as soon as i activated our integration again, it stops working and is timing out again. 😞 A power cycle also makes it work for a couple of minutes, and then it stops again. 

What do you guys think is happening?

I thought maybe we are hammering it to hard, but then again the would be weird if i am the only one having this problem.

## Version of the integration

If you are unsure about the version check the [github release here](https://github.com/leikoilja/ha-google-home/releases).

```text
1.4.2
```

## Logs

```text
2021-04-13 19:54:40 DEBUG (MainThread) [custom_components.google_home] Fetching data from Google Home device Køkken - https://192.168.0.205:8443/setup/assistant/alarms
2021-04-13 19:54:42 DEBUG (MainThread) [custom_components.google_home] Køkken device timed out while trying to get alarms and timers.
2021-04-13 19:54:42 DEBUG (MainThread) [custom_components.google_home] Getting Do Not Disturb setting from Google Home device Køkken
2021-04-13 19:54:42 DEBUG (MainThread) [custom_components.google_home] Requesting endpoint setup/assistant/notifications for Google Home device Køkken - https://192.168.0.205:8443/setup/assistant/notifications
2021-04-13 19:54:44 DEBUG (MainThread) [custom_components.google_home] Køkken device timed out while trying to post data to it - Raw data: None
2021-04-13 19:54:44 DEBUG (MainThread) [custom_components.google_home] Finished fetching sensor data in 4.009 seconds
```

KapJI:
I had something similar with Nest Hub Mini yesterday but then it remediated itself. Seems to work since then.

leikoilja:
I did have the same behavior as well a few times. Thou currently my RPi has just stopped working with constant "Unable to connect to Home Assistant" since morning. Nothing seems to be helping, so i ll just focus on fixing it first 😅

DurgNomis-drol:
[...]

PR #302, opened 2021-08-10
Add config option for changing update interval.

DurgNomis-drol:
As discussed in #202 and [comment](https://github.com/leikoilja/ha-google-home/issues/202#issuecomment-895221500)

Adds ability for the user to change the update interval. This will make it possible for users who suffer for timeout problems to change their update interval as it seems this can help with it.

Also this now sets the default update interval to 180 seconds instead of 10 as pointed out in comment. Users should at their own risk tune it what they need.

This does not close. #202 but gives users more options to try to mitigate it.

### DIFF
--- a/custom_components/google_home/__init__.py
+++ b/custom_components/google_home/__init__.py
@@ -20,6 +20,7 @@ from .api import GlocaltokensApiClient
 from .const import (
     CONF_ANDROID_ID,
     CONF_MASTER_TOKEN,
+    CONF_UPDATE_INTERVAL,
     DATA_CLIENT,
     DATA_COORDINATOR,
     DOMAIN,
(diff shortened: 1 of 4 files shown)


leikoilja (approved):
Looks good, thanks for fixing it!

Were you able to validate it's working locally? :) 
It would be so so nice to add a few lines to "README" about integration limitations explaining that the fetch rate is 3min and the user can tune it at their own risk. If we add it in the first part of README it can save a lot of time to some people. What do you think? (Edit: just saw your other comment that you were thinking of making a README as a separate PR. Sounds great 🚀 )

DurgNomis-drol:
Thanks for approving🎉 I tested it locally and it works for me. 😊

KapJI (approved):
Looks good, thanks!

leikoilja:
@DurgNomis-drol before we can release it in a next public release, can you please draft your thoughts regarding limitations in `README`? :)

DurgNomis-drol:
Yes, I will try to get it done later today 😊

PR #310, opened 2021-08-16
Using unique device_id received from glocaltokens==0.4.0

### LINKED ISSUE #199: Feature - Change unique id from device_name to something really unique.
We can either use [this](https://rithvikvibhu.github.io/GHLocalApi/#operation/EurekaInfo) endpoint or as @leikoilja mentioned, we can generate one in glocaltokens.

What do you guys think is the best option?

**Original:**
Right now we use `device_name` in entity unique id. I think glocaltokens can provide some real device specific id for the device. Then HA should handle renames itself.

_Originally posted by @KapJI in https://github.com/leikoilja/ha-google-home/issues/184#issuecomment-818705358_

leikoilja:
`glocaltokens` v0.4.0 [implements](https://github.com/leikoilja/glocaltokens/pull/145) grabbing unique device id from foyer API 

This PR is utilising that `unique_id` for device id.
Closes #199 

This is probably a breaking change since it will require users to reload the integration so the devices receive new unique ID.
Note, after this is merged into `master` we need to extensively test it before releasing a new public release.

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -92,6 +92,7 @@ class GlocaltokensApiClient:
             google_devices = await self.hass.async_add_executor_job(_get_google_devices)
             self.google_devices = [
                 GoogleHomeDevice(
+                    device_id=device.device_id,
                     name=device.device_name,
                     auth_token=device.local_auth_token,
                     ip_address=device.ip_address,
--- a/custom_components/google_home/entity.py
+++ b/custom_components/google_home/entity.py
@@ -21,10 +21,12 @@ class GoogleHomeBaseEntity(CoordinatorEntity, ABC):
         self,
         coordinator: DataUpdateCoordinator,
         client: GlocaltokensApiClient,
+        device_id: str,
         device_name: str,
     ):
         super().__init__(coordinator)
         self.client = client
+        self.device_id = device_id
         self.device_name = device_name
 
     @property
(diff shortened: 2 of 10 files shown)


DurgNomis-drol:
I can review and test later 😊

DurgNomis-drol (approved):
LGTM! 🎉 

I just tested your local branch and it works for me. 💯 

After looking at the unique id, it looks like it a combination of mac and some other info. So i think it is safe to use.

leikoilja:
nice! thanks for checking it, @DurgNomis-drol

leikoilja:
@DurgNomis-drol, i have just pulled `master` version of integration and discovered that it has created duplicate sensors for me (as expected), although the `unique_id` names are not what I expected. 🤔  I was expecting `device_id` to be used for `unique_id` 🤔 
image
Do you have the same? 

Also drafted some breaking change warnings in release drafts for next release 
image

DurgNomis-drol:
`entity_id` (as you have in the dialog box) is not the same as `unique_id` or `device_id` in HA.
So this is in fact expected behavior, and also what we want. We are now sure that every device, even if it have the same name, is uniquely identified as different devices in HA device registry. When a unique_id is set, it is possible to change the `name` and `entity_id` in the UI, but it is still linked to the `unique_id` / `device_id`.

The unique_id is something like this `d42efd43-1518-4a59-bdb1-78cf4ab4fbb9`.

leikoilja:
ou, haha, clearly i confused the difference between `unique_id` and `entity_id` on the device card (screenshot above), I guess I was expecting to see the `unique_id` there :D 
Thanks for laying it down, @DurgNomis-drol. Now that we both have tested it - we good to release? 🚀

DurgNomis-drol:
@leikoilja haha that's okay, it took me sometime to also understand the difference between them 😁

Yes I think so!

PR #322, opened 2021-09-05
Match devices by device_id

KapJI:
If device is renamed or appears on homegraph under different name, it still should be matched by corresponding entity.


leikoilja (approved):
Superb 🚀

PR #323, opened 2021-09-05
Better update handler

KapJI:
- Instead of reloading the whole integration, just set new `update_interval` in coordinator.
- Use new helper functions from config_entry to reduce boilerplate code.
- Log successful response, can be useful for debugging.
- Show codes of mypy errors.

### DIFF
--- a/custom_components/google_home/api.py
+++ b/custom_components/google_home/api.py
@@ -167,14 +167,14 @@ class GlocaltokensApiClient:
                 device.set_timers(cast(List[TimerJsonDict], response[JSON_TIMER]))
                 device.set_alarms(cast(List[AlarmJsonDict], response[JSON_ALARM]))
                 _LOGGER.debug(
-                    "Successfully retrieved data from %s. Response: %s",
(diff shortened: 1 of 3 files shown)


custom_components/google_home/api.py
```diff
@@ -166,7 +166,11 @@ async def update_alarms_and_timers(
             if JSON_TIMER in response and JSON_ALARM in response:
                 device.set_timers(cast(List[TimerJsonDict], response[JSON_TIMER]))
                 device.set_alarms(cast(List[AlarmJsonDict], response[JSON_ALARM]))
-                _LOGGER.debug("Successfully retrieved data from %s.", device.name)
+                _LOGGER.debug(
+                    "Successfully retrieved data from %s. Response: %s",
```
leikoilja, line 170:
```suggestion
                    "Successfully retrieved data alarms and timer data from %s. Response: %s",
```

leikoilja (approved):
This is amazing. I m happy someone got at looking for optimisation of HA related logic 💥 
Thanks yoy

PR #324, opened 2021-09-05
Remove data collection option

KapJI:
Data collection is not implemented and I don't see any value in having unused option.

It turns out user reports and logs are sufficient for debugging.

Can be reverted once if at some point we decide to implement it.


DurgNomis-drol (approved):
LGTM! 🎉

leikoilja (approved):
Thanks for cleaning it up, I was thinking of the same. 
We had a discussion about data collection in https://github.com/leikoilja/ha-google-home/discussions/116 but never got around to implement it.
Let's remove this option for now as we might not even use it :)

PR #325, opened 2021-09-05
Workflow for publishing new releases

KapJI:
Use Github API to fetch release draft generated by release-drafter, commit updated `manifest.json` and publish that release.

It can be run manually from Actions tab. 
Draft can be modified before running it.

I've tested only parts of it, so no guarantee that it'll work on the first try.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,34 @@
+name: Publish new release
+
+on:
+  workflow_dispatch:
+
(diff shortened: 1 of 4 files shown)


leikoilja (approved):
This is soooo freaking cool 💯  :O 
i am stocked about all of that automation with publishing new releases 🤩
Feel free to merge in the other approved PRs and release a new version with this amazing action. Thanks thanks thanks 🙌

PR #328, opened 2021-09-06
Temporary disable branch protection for updating manifest

KapJI:
Disable branch protection rules for admin to commit updated `manifest.json`.


leikoilja (approved):
Nice. Do we need ADMIN token for that? 🤔

KapJI:
Yes, I think so. Github actions token is not admin and can't modify protection rules.

leikoilja:
makes sense. Let's keep trying, but for the heads up the current token is just living 30 days, so when it's expired i ll create another one that we can use :) Still not sure about the safety of it, but dont think there are better options yet

PR #329, opened 2021-09-06
Do not format manifest.json with prettier

KapJI:
It turns out prettier formats JSON differently than `json.dumps`: https://github.com/leikoilja/ha-google-home/runs/3524470666

Let's only check that JSON is valid with `check-json` hook and exclude `manifest.json` from formatting.
It'll be rewritten by publishing script on every release anyway.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,4 +1,3 @@
-// See https://aka.ms/vscode-remote/devcontainer.json for format details.
 {
   "image": "ludeeus/container:integration",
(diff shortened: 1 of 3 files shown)


KapJI:
@leikoilja it seems branch protection rule is disabled, can you check it please?

leikoilja (approved):
Nice, thank you.  💥

Also, weirdly enough the branch protection rule was disabled right now. I m pretty sure I have not done that manually, so let's keep an eye on that after the next automated release. Might be that the bot disables it and fails to re-enable :)

PR #332, opened 2021-09-09
Correctly change admin enforcement in publish script

KapJI:
`edit_protection()` actually rewrites all settings and not setting them clears them out.

This should preserve current settings and change only admin enforcement bit.


leikoilja (approved):
Nice! Thanks man 💥

PR #366, opened 2021-10-16
Add danish translation

DurgNomis-drol:


DurgNomis-drol:
Something is strange with CI, it shouldn't take this long to complete...

Issue #359, opened 2021-10-08
Volume is not a percentage


TheFes:
## Describe the bug

the new number entity shows it's value with unit_of_measure %, but it not a percentage, as 1 is the maximum value. (this should be 100% if it was indeed a percentage)

## Version of the integration

If you are unsure about the version check the [github release here](https://github.com/leikoilja/ha-google-home/releases).

```text
1.82.2
```

## Logs

```text
none relevant
```

leikoilja:
Fair point, thanks for reporting @TheFes. 
Noticed that the other day and didn't want to change it yet cause it will become a breaking change for some who already have automations tuned to use these values. But we need to fix that and release a minor version with breaking change notes 🙃 will mark it as a bug so we can more it further

TheFes:
You could also just remove the unit_of_measure.

PR #371, opened 2021-10-20
Use round instead of int

DurgNomis-drol:
Fixes https://github.com/leikoilja/ha-google-home/issues/359#issuecomment-947511583

Should we also round the float to less decimals?


DurgNomis-drol:
@leikoilja Haha 😆 No problem

PR #387, opened 2021-11-10
Clarify that Logger service needs to be enabled & configured in yaml mode

jlsjonas:
Logger: Set level only appears if `logger:` is defined in the configuration.yaml


ArnyminerZ (approved):

leikoilja:
Awesome, thanks @jlsjonas and @ArnyminerZ 🚀 🔥

PR #392, opened 2021-11-14
Move attributes to device model and remove integration attribute

DurgNomis-drol:
Just a bit of cleanup 😄 

- Moves `hardware` attribute to `device_info`.
- Removes `integration` attribute as the integration is specified in `device_info`
- Change `MANUFACTURER` to human-readable instead as this is the name shown in the UI and should not be `google_home` but Google Home.


DurgNomis-drol:
Tested and works locally

leikoilja (approved):
Looks good to me 💥 
Not sure what `integration` could have been used in device attributes and why we had it there to begin with, but this cleanup looks good :) 
Thanks 💯

Issue #402, opened 2021-11-28
Failed to fetch timers/alarms information from device


BossHuso:
## Describe the bug

I have everything set up as described and succesfully integrated within HA. But I can't get my devices.

One important note perhaps, is that HA is running on a server **not** within my home network. I saw some other issues where the logs of the Google devices return a local IP. If this is the case, I get that it won't work. But since my Google Home app can access the devices from every network, I thought it should be possible with this integration as well?

## Version of the integration

If you are unsure about the version check the [github release here](https://github.com/leikoilja/ha-google-home/releases).

```text
Latest pulled from git, so I'm guessing that counts as Release v1.9.1
```

## Logs

```text
2021-11-28 16:13:20 DEBUG (MainThread) [custom_components.google_home] Failed to fetch timers/alarms information from device Zara's kamer. We could not determine its IP address, the device is either offline or is not compatible Google Home device. Will try again later.
2021-11-28 16:13:20 DEBUG (MainThread) [custom_components.google_home] Failed to fetch timers/alarms information from device Babykamer. We could not determine its IP address, the device is either offline or is not compatible Google Home device. Will try again later.
2021-11-28 16:13:20 DEBUG (MainThread) [custom_components.google_home] Failed to fetch timers/alarms information from device Woonkamer. We could not determine its IP address, the device is either offline or is not compatible Google Home device. Will try again later.
2021-11-28 16:13:20 DEBUG (MainThread) [custom_components.google_home] Failed to fetch timers/alarms information from device TV. We could not determine its IP address, the device is either offline or is not compatible Google Home device. Will try again later.
```

DurgNomis-drol:
You would need to set up routing between each network, but that is out of scope for this integration. 
The reason that the Google Home app can see your devices is because it can connect to them through the cloud. This integration communicates directly with each device. We only connect to Google servers to authenticate and get the auth token for each device. Besides that, we use mdns to discover devices (There IP address). That will also not work if not connected to the same network or not routed correctly between each. I believe other people have had similar issues with HA (not this integration) and you can probably find more info on how to set it up on the forums. If set up correctly, this integration should work just fine. Hope this helps you in the right direction. 😊

BossHuso:
[...]

PR #403, opened 2021-11-29
Add explanation for server and devices not on the same network

ArnyminerZ:
Added an explanation on README for having some reference over #402 and informing that the device and the Home Assistant installation must be on the same network.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ If you get the following error for ***all*** of your devices:
 2021-11-28 16:13:20 DEBUG (MainThread) [custom_components.google_home] Failed to fetch timers/alarms information from device xxx. We could not determine its IP address, the device is either offline or is not compatible Google Home device. Will try again later.
 ```
 
-I may be that your device and Home Assistant's installation are not on the same network.
+It may be that your device and Home Assistant installation are not on the same network.
 
 The integration works by connecting to the Google's servers to authenticate and get the authorisation keys for controlling these devices, but after that, all the requests are made locally, so it's required that the server and devices are on the same network. You can use a VPN or setup routing between each network to overcome this issue.
 
--- a/README.md
+++ b/README.md
@@ -377,7 +377,7 @@ If you get the following error for ***all*** of your devices:
 
 I may be that your device and Home Assistant's installation are not on the same network.
 
-The integration works by connecting to the Google's servers to authenticate and get the authorisation keys for controlling the devices, but after that, all the requests are made locally, so it's required that the server and devices are on the same network. You can use a VPN to overcome this issue.
+The integration works by connecting to the Google's servers to authenticate and get the authorisation keys for controlling these devices, but after that, all the requests are made locally, so it's required that the server and devices are on the same network. You can use a VPN or setup routing between each network to overcome this issue.
 
 ## Contribution
 
--- a/README.md
+++ b/README.md
@@ -367,6 +367,16 @@ If unsure, please check what account you are using in the _Google Home_ app and 
 
 If you can see your devices, and they all seem correct, but the alarms/timers sensors do not appear, or appear empty try restarting the Google Home device, Home Assistant, and reinstalling the integration.
 
+### Device offline or not compatible message (#402)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -367,6 +367,18 @@ If unsure, please check what account you are using in the _Google Home_ app and
 
 If you can see your devices, and they all seem correct, but the alarms/timers sensors do not appear, or appear empty try restarting the Google Home device, Home Assistant, and reinstalling the integration.
 
+### Device offline or not compatible message (#402)
+
+If you get the following error for all of your devices:
```
DurgNomis-drol, line 372: Can we make this bold and italic so it catches peoples eyes?

```suggestion
If you get the following error for ***all*** of your devices:
```

README.md
```diff
@@ -367,6 +367,18 @@ If unsure, please check what account you are using in the _Google Home_ app and
...
+
+```
+2021-11-28 16:13:20 DEBUG (MainThread) [custom_components.google_home] Failed to fetch timers/alarms information from device xxx. We could not determine its IP address, the device is either offline or is not compatible Google Home device. Will try again later.
+```
+
+I may be that your device and Home Assistant's installation are not on the same network.
+
+The integration works by connecting to the Google's servers to authenticate and get the authorisation keys for controlling the devices, but after that, all the requests are made locally, so it's required that the server and devices are on the same network. You can use a VPN to overcome this issue.
```
DurgNomis-drol, line 380:
```suggestion
The integration works by connecting to the Google's servers to authenticate and get the authorisation keys for controlling these devices, but after that, all the requests are made locally, so it's required that the server and devices are on the same network. You can use a VPN or setup routing between each network to overcome this issue.
```

DurgNomis-drol (approved):
LGTM :rocket: Please see my small comments :smiley:

README.md
```diff
@@ -367,6 +367,18 @@ If unsure, please check what account you are using in the _Google Home_ app and
...
 
+### Device offline or not compatible message (#402)
+
+If you get the following error for all of your devices:
+
+```
+2021-11-28 16:13:20 DEBUG (MainThread) [custom_components.google_home] Failed to fetch timers/alarms information from device xxx. We could not determine its IP address, the device is either offline or is not compatible Google Home device. Will try again later.
+```
+
+I may be that your device and Home Assistant's installation are not on the same network.
```
KapJI, line 378:
```suggestion
It may be that your device and Home Assistant installation are not on the same network.
```

README.md
KapJI:
```suggestion
The integration works by connecting to the Google servers to authenticate and get the authorisation keys for the devices, but after that, all the requests are made locally, so it's required that the server and devices are on the same network. You can use a VPN to overcome this issue.
```

DurgNomis-drol: @KapJI I think it is important to include the note i added, that it can also be achieved by setting up routing between each network, in case people are using VLAN's.

KapJI: I'm not against it, my changes are for the current version.

README.md
```diff
@@ -367,6 +367,18 @@ If unsure, please check what account you are using in the _Google Home_ app and
 
 If you can see your devices, and they all seem correct, but the alarms/timers sensors do not appear, or appear empty try restarting the Google Home device, Home Assistant, and reinstalling the integration.
 
+### Device offline or not compatible message (#402)
+
+If you get the following error for ***all*** of your devices:
```
leikoilja, line 372: Based on the failed linting: https://github.com/leikoilja/ha-google-home/runs/4373080559?check_suite_focus=true

```suggestion
If you get the following error for **_all_** of your devices:
```

leikoilja (approved):
Awesome stuff! Thanks @ArnyminerZ for adding this info to trouble shooting guide :) I wonder if we should soon open a separate docs page (github wiki or some other type of documentation system) where we move all the knowledge base and keep README nice and clean with just essential information? :)

DurgNomis-drol:
@leikoilja I like that idea, was just thinking the same! I believe that we can setup a wiki directly in this repo.

PR #414, opened 2021-12-10
device_state_attributes warnings 2021.12.0b

covid10:
Fix for device_state_attributes warnings which began in release 2021.12.0b


DurgNomis-drol (approved):
LGTM!

Thanks for contributing ❤️

Issue #420, opened 2021-12-22
5 is not a valid GoogleHomeAlarmStatus


philippefs35:
## Describe the bug

Integration fails to load due to an unknown alarm status
![image](https://user-images.githubusercontent.com/1505656/147113937-3abb3962-6bcf-42a6-a289-058ac5a9434c.png)

## Version of the integration

```text
1.9.3
```

## Logs

```text
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 187, in _async_refresh
    self.data = await self._async_update_data()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 147, in _async_update_data
    return await self.update_method()
  File "/config/custom_components/google_home/api.py", line 139, in update_google_devices_information
    coordinator_data = await asyncio.gather(
  File "/config/custom_components/google_home/api.py", line 152, in collect_data_from_endpoints
    device = await self.update_alarms_and_timers(device)
  File "/config/custom_components/google_home/api.py", line 168, in update_alarms_and_timers
    device.set_alarms(cast(List[AlarmJsonDict], response[JSON_ALARM]))
  File "/config/custom_components/google_home/models.py", line 48, in set_alarms
    self._alarms = [
  File "/config/custom_components/google_home/models.py", line 49, in <listcomp>
    GoogleHomeAlarm(
  File "/config/custom_components/google_home/models.py", line 169, in __init__
    self.status = GoogleHomeAlarmStatus(status)
  File "/usr/local/lib/python3.9/enum.py", line 384, in __call__
    return cls.__new__(cls, value)
  File "/usr/local/lib/python3.9/enum.py", line 702, in __new__
    raise ve_exc
ValueError: 5 is not a valid GoogleHomeAlarmStatus
```

KapJI:
What alarms do you have set?

How it looks like in the Home app on this screen?
![image](https://user-images.githubusercontent.com/3767331/147119536-94fe40e0-f61c-4eab-995d-2707e7ee6942.jpeg)

philippefs35:
So I had the alarm in this screenshot (7:30 Mon, Tue, Wed, Thu, Fri), but it appeared twice:
![image](https://user-images.githubusercontent.com/1505656/147132501-7be715f6-fbfe-4a09-8c95-0b31415e50cc.png)

That is: the exact same line appeared twice (even though it only appeared once on my Nest Hub)

[...]

PR #428, opened 2022-01-04
Added new missed state for alarms (#420)

williamg97:
Added status 5 which seems to be missed status. I was able to narrow it down to which of my many devices was causing it and then could see this alarm missed alert when clicking on the alarm icon. Once I'd cleared the following it started to work again. 

image

### DIFF
--- a/custom_components/google_home/sensor.py
+++ b/custom_components/google_home/sensor.py
@@ -179,7 +179,9 @@ class GoogleHomeAlarmsSensor(GoogleHomeBaseEntity):
         next_alarm = device.get_next_alarm()
         return (
             next_alarm.local_time_iso
-            if next_alarm and next_alarm.status != GoogleHomeAlarmStatus.INACTIVE
+            if next_alarm
+            and next_alarm.status != GoogleHomeAlarmStatus.INACTIVE
+            and next_alarm.status != GoogleHomeAlarmStatus.MISSED
             else STATE_UNAVAILABLE
         )
 
--- a/custom_components/google_home/sensor.py
+++ b/custom_components/google_home/sensor.py
@@ -180,8 +180,7 @@ class GoogleHomeAlarmsSensor(GoogleHomeBaseEntity):
         return (
             next_alarm.local_time_iso
             if next_alarm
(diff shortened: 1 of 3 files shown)


custom_components/google_home/sensor.py
```diff
@@ -179,7 +179,9 @@ def state(self) -> str | None:
         next_alarm = device.get_next_alarm()
         return (
             next_alarm.local_time_iso
-            if next_alarm and next_alarm.status != GoogleHomeAlarmStatus.INACTIVE
+            if next_alarm
+            and next_alarm.status != GoogleHomeAlarmStatus.INACTIVE
+            and next_alarm.status != GoogleHomeAlarmStatus.MISSED
```
leikoilja, line 184: How about a shorthand version here to keep the implementation consistent?

```suggestion
            and next_alarm.status not in (GoogleHomeAlarmStatus.INACTIVE, GoogleHomeAlarmStatus.MISSED)
```

leikoilja (approved):
Amazing! Thanks, @williamg97 for digging out and finding what missing status 5 means :)

custom_components/google_home/sensor.py
```diff
@@ -179,7 +179,8 @@ def state(self) -> str | None:
         next_alarm = device.get_next_alarm()
         return (
             next_alarm.local_time_iso
-            if next_alarm and next_alarm.status != GoogleHomeAlarmStatus.INACTIVE
+            if next_alarm
+            and next_alarm.status not in (GoogleHomeAlarmStatus.INACTIVE, GoogleHomeAlarmStatus.MISSED)
```
DurgNomis-drol, line 183:
```suggestion
            and next_alarm.status
            not in (GoogleHomeAlarmStatus.INACTIVE, GoogleHomeAlarmStatus.MISSED)
```

DurgNomis-drol (changes requested):

leikoilja (approved):

leikoilja:
Awesome find, @williamg97! Thanks for fixing it 🚀 🔥

PR #451, opened 2022-02-03
Add a reminder to use github action for releases

KapJI:
Change title to remind about that, draft should be visible only to collaborators. Script will then cut this message from the final release.


leikoilja (approved):
haha awesome tweak! thanks for adding it, will surely help :)

Issue #470, opened 2022-02-24
2022.03.0b0 fails to load


brianhanifin:
## Describe the bug

After installing Home Assistant 2022.03.0b0 this integration fails to load with the following error  

```text
Logger: homeassistant.setup
Source: setup.py:155 
First occurred: 4:47:43 PM (1 occurrences) 
Last logged: 4:47:43 PM

Setup failed for google_home: Requirements for google_home not found: ['glocaltokens==0.6.2'].
```

## Version of the integration

If you are unsure about the version check the [github release here](https://github.com/leikoilja/ha-google-home/releases).

```text
1.9.7
```

## Logs

```text
Add your logs here.
```

pschneider87:
yes, same here with latest beta release

ArnyminerZ:
Try to install `glocaltokens` manually with
```
pip3 install glocaltokens
```
And paste the command feedback.

leikoilja:
Seems to be caused by Python pips newer resolver, see https://github.com/leikoilja/glocaltokens/issues/218 for more info.

leikoilja:
Fixed and released in [v1.9.8](https://github.com/leikoilja/ha-google-home/releases/tag/v1.9.8) 💥

pschneider87:
I can't install. Hacs times out when installing latest version

leikoilja:
hmmm, sound a bit more like HACS issue? Does it still happen, @pschneider87 ?

pschneider87:
![Screenshot_20220226-155525~2.png](https://user-images.githubusercontent.com/17707309/155847674-d9a19f9a-a824-4f5f-a303-93d171e2276f.png)

Yes, I still get the error when try to update

leikoilja:
@pschneider87, from the error extract you have sent above it, seems that HA is timeouting while installing the integration. Can you please first check your network connection on your HA (as the log suggests)? Secondly, make sure you are on the latest versions of HACS / Home Assistant OS / Home Assistant. Thirdly, make sure you remove the old integration (just to avoid any conflicts. And at last, if none of the previous steps have helped, please open a new issue filling up all required information as well as logs extracts, so we could have more information how to help you :)

pschneider87:
seems to be an HACS issue, can't reinstall, and other integrations are failing to update as well.

Issue #467, opened 2022-02-21
Configuration flow could not be loaded: [Object Object]


Timo2208:
Config flow kon niet geladen worden: [object Object]

Configuration flow could not be loaded: [Object Object]

did add the debug view but dont get the log files's

when adding the integratin its giving this message

leikoilja:
What version of the home assistant, what version of the integration and what version of HACS are you trying to use? What system are you on?

leikoilja:
Can confirm the issue. Now running the integration in the local dev container I can see:
<img width="852" alt="image" src="https://user-images.githubusercontent.com/10655107/155582447-97b7924b-fd8e-4007-af2f-a72a50f592f4.png">

in logs when trying to install `Google Home` integration

leikoilja:
Fixed and released in [v1.9.8](https://github.com/leikoilja/ha-google-home/releases/tag/v1.9.8) 💥

PR #472, opened 2022-02-24
Bump glocaltokens to 0.6.3 and adding google-api-python-client dep

leikoilja:
Bumps glocaltokens to 0.6.3 (that fixes the dependency issues)
and add google-api-python-client dep.

@KapJI, maybe you have some good thoughts about it. SInce latest HA is using pip v22.0 which has it's revolver changed we have dep issues (that's the reason why glocaltokens was failing to install to HA). However now that glocaltokens got fixed during the integration installation I have got an unexpected error that `google` dependency is missing:
image
With the same UI error message `Configuration flow could not be loaded: [Object Object]` as in https://github.com/leikoilja/ha-google-home/issues/467

A quick search has led me to [this](https://stackoverflow.com/a/53901668/5818549), so adding a new dependency `google-api-python-client` does indeed fix the issue. Could that be right? I don't see how it's fighting together. Or maybe we should move this `google-api-python-client` dependency to `glocaltokens` instead? 

Closes https://github.com/leikoilja/ha-google-home/issues/467
Closes https://github.com/leikoilja/ha-google-home/issues/470

### DIFF
--- a/custom_components/google_home/manifest.json
+++ b/custom_components/google_home/manifest.json
@@ -19,6 +19,7 @@
   "requirements": [
     "glocaltokens==0.6.2"
   ],
+  "requirements": ["glocaltokens==0.6.3", "google-api-python-client==2.38.0"],
   "version": "1.9.7",
   "zeroconf": [
     "_googlecast._tcp.local."
--- a/custom_components/google_home/manifest.json
+++ b/custom_components/google_home/manifest.json
@@ -20,7 +20,7 @@
     "glocaltokens==0.6.2"
   ],
   "requirements": [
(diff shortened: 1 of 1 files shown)


KapJI (approved):
Please revert the formatting to keep the blame. 

Regarding issue with google import, adding dependency sounds right. I'm away from computer and can't check where is the right place to add it. It should be added where it's being used.

stboch:
Tested the changes with 2022.3.b1 and it resolved the issues successfully.

leikoilja:
Fixed and released in [v1.9.8](https://github.com/leikoilja/ha-google-home/releases/tag/v1.9.8)

PR #490, opened 2022-03-21
Use permanent link for installs badge

KapJI:
It'll be redirected to the current endpoint.


leikoilja (approved):
Oh that's fancy. Did you register an account at `runkit` and linked it to yourself, @KapJI?

KapJI:
I created this endpoint it in my account in the first place. It seems we should use the link before the redirect. The link after redirect (with a hash) may change over time while this one should be permanent.

KapJI:
Can you open it directly? Maybe the image is still cached?
https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Frunkit.io%2Fkapji%2Fgoogle-home-installs-for-shield-io%2F3.0.0

leikoilja:
yeh, the link straight works perfectly fine :) let's give it a little moment to update, will probs be fine in a sec

PR #504, opened 2022-04-01
Use EntityCategory enum instead of strings

### LINKED ISSUE #503: Error with latest Beta core 2022.4.0b1
## Describe the bug

image

Feedback from Discord suggests the following "That looks like custom components using the string instead of the enum for EntityCategory"

## Version of the integration

[v1.9.9]

```text
Home Assistant 2022.4.0b1
```

## Logs

```text
2022-04-01 10:32:48 ERROR (MainThread) [homeassistant.components.switch] Error adding entities for domain switch with platform google_home
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 382, in async_add_entities
    await asyncio.gather(*tasks)
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 510, in _async_add_entity
    entry = entity_registry.async_get_or_create(
  File "/usr/src/homeassistant/homeassistant/helpers/entity_registry.py", line 345, in async_get_or_create
    return self.async_update_entity(
  File "/usr/src/homeassistant/homeassistant/helpers/entity_registry.py", line 529, in async_update_entity
    raise ValueError("entity_category must be a valid EntityCategory instance")
ValueError: entity_category must be a valid EntityCategory instance

```

mdegat01:
Using the entity category strings doesn't work anymore starting in `2022.4`. Using the enums instead fixes #503 .

The enums have been accepted since `2021.12` ([PR](https://github.com/home-assistant/core/pull/60720)) which covers your supported versions so it is safe to make immediately, don't have to wait for `2022.4`.

### DIFF
--- a/custom_components/google_home/number.py
+++ b/custom_components/google_home/number.py
@@ -5,8 +5,9 @@ import logging
 
 from homeassistant.components.number import NumberEntity
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import ENTITY_CATEGORY_CONFIG, PERCENTAGE
+from homeassistant.const import PERCENTAGE
 from homeassistant.core import HomeAssistant
+from homeassistant.helpers.entity import EntityCategory
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator
 
@@ -57,7 +58,7 @@ class AlarmVolumeNumber(GoogleHomeBaseEntity, NumberEntity):
     """Google Home Alarm Volume Number entity."""
 
     _attr_unit_of_measurement = PERCENTAGE
-    _attr_entity_category = ENTITY_CATEGORY_CONFIG
+    _attr_entity_category = EntityCategory.CONFIG
 
     @property
     def label(self) -> str:
--- a/custom_components/google_home/sensor.py
+++ b/custom_components/google_home/sensor.py
@@ -8,12 +8,11 @@ import voluptuous as vol
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
     DEVICE_CLASS_TIMESTAMP,
-    ENTITY_CATEGORY_DIAGNOSTIC,
     STATE_UNAVAILABLE,
 )
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers import config_validation as cv, entity_platform
-from homeassistant.helpers.entity import Entity
+from homeassistant.helpers.entity import Entity, EntityCategory
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 
 from .const import (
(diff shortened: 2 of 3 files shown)


ArnyminerZ (approved):

leikoilja (commented):
That's nice, thank you! 
Are we sure it's backwards competitive and won't break the integration for those using latest stable HA image? 🤔  :)

ArnyminerZ:
> Are we sure it's backwards competitive and won't break the integration for those using latest stable HA image? 🤔 :)

Well, I don't think so. `EntityCategory` is added in home-assistant/core#60720, so all compilations before that one will not contain that class, resulting on an error, as far as I see it. Is there any way to check the version, and choose accordingly?

If not, maybe we should wait until the stable version is released, and then use the new method, which won't be compatible with HA `<2022.4`.

However, this should be tested.

ArnyminerZ:
After a bit of searching I've found [this](https://stackoverflow.com/a/3496790/5717211) on StackOverflow, so maybe something like this can be used:
```python3
if version >= "2022.4":
    from homeassistant.const import ENTITY_CATEGORY_CONFIG
else:
    from homeassistant.helpers.entity import EntityCategory.CONFIG as ENTITY_CATEGORY_CONFIG
```
Which will allow to keep using
```python3
_attr_entity_category = ENTITY_CATEGORY_CONFIG
```
Not sure about this, it has to be tested and further investigation must be performed, just a suggestion that might work out until it's marked as not backward-compatible by HA.

mdegat01:
@ArnyminerZ This repository lists version `2022.2` as it's minimum supported version of home assistant which is after the linked pr.
https://github.com/leikoilja/ha-google-home/blob/master/hacs.json#L4

That's why I said it would be fine to merge immediately. All versions of home assistant this integration supports are after that was merged into core and the enum shipped.

I'll fix the formatting and stuff though sorry about that. Was cranking through a couple of these based on users reporting the same issues in the beta channel of ha

mdegat01:
Also just to get everyone on the same page, 2022.4 is the release when providing an entitycategory as a string instead of an enum goes from being deprecated to being unsupported. It's **not** the release when HA introduced the entitycategory enum, that was 2021.12 (the pr we both linked was committed into core on 12/1/2021).

So essentially this component no longer works at all starting in 2022.4 without this change. With this change it works fine for all releases of ha >= 2021.12 which covers all supported releases of this component according to it's HACS.json.

ArnyminerZ:
Okey @mdegat01, thanks for making it clear.  Then we can merge this without any issues. Agree @leikoilja ?

ArnyminerZ:
Since no further issues have been reported, merging this one 🚀

grantclem:
This issue does not appear to be resolved, check https://github.com/leikoilja/ha-google-home/issues/503 for update

leikoilja:
Awesome, thank you, gents 💪 💥 🔥

grantclem:
Thanks very much for your efforts resolving this.

Issue #339, opened 2021-09-17
Allow adding integration with master token alone


leikoilja:
I think it would be a useful option to allow the user to specify a `master_token` instead of the account username and app password as is suggested as best security practice in the `glocaltokens` readme. Having to store my Google account credentials on the home assistant server is the main thing keeping me from using this extension so any way around that is appreciated, this was just what I came up with to maybe achieve that.

_Originally posted by @coleya in https://github.com/leikoilja/ha-google-home/issues/124#issuecomment-921531375_

Related https://github.com/leikoilja/glocaltokens/issues/168

DurgNomis-drol:
Should we rethink how (and if should) we store the `master_token`? HA does not encrypt anything AFAIK, but i am no security expert. What do you guys think?

And by storing, I mean not saving the `master_token` in config flow, but rather get one each time the integration is loaded from username/password? 🤔

leikoilja:
@DurgNomis-drol, I would love to and I think it's very important, but tbh, i don't know what security considerations can we do within our integration scope. Currently we store username, password, master_token and android_id in the integration instance. But so does all other HA integrations, leaving the rest of security vulnerabilities/encryption(?)/storage etc for HA to manage. Apart from that "pre-defined" flow I am not sure what we can do 🤷‍♂️

DurgNomis-drol:
@leikoilja You are probably right, i was just thinking out loud after the back and forth in the other issue. 😃

leikoilja:
Totally agree with you. We can still support master token as an auth method, thou i dont think it's a common use case and it's not going to make it easier to authenticate or use the integration. 
Let's also hear from @KapJI and @Drakulix, maybe they have some thoughts on security and better practises :)

KapJI:
I think encrypting any data is meaningless because it should be decrypted by HA, and the decryption key must be stored somewhere where HA can access it. If someone has access to these files you're pwned in any case, it doesn't matter if they're encrypted or not.

Not storing master token doesn't make any sense either if we still store username and password. Obtaining it every time should work but the only thing it'll do is making the initialization of the integration slower because of the extra request.

Allowing using master token in the config flow is ok if we can provide instructions on how to obtain it. Although I don't think this makes anything safer, probably it's only a bit easier to revoke such token.

Drakulix:
> Let's also hear from @KapJI and @Drakulix, maybe they have some thoughts on security and better practises :)

[...]

PR #511, opened 2022-04-11
Create security-report.md

ArnyminerZ:
We should provide a way for users to report security issues #339 would be a good example of this type of reports.


ArnyminerZ:
Yeah, I agree this won't be common, but just in case, it's good to have.

PR #555, opened 2022-06-27
Update dependencies

KapJI:
- Remove `google-api-python-client`. It doesn't belong here. It was solving problem with `protobuf` not installed as a transitive dependency of `grpcio` on some systems. Protobuf is used by `glocaltokens` and now it has explicit dependency to `protobuf`. This fix only worked by accident, because `google-api-python-client` transitively depends on `protobuf`:
```
google-api-python-client 2.51.0 Google API Client Library for Python
├── google-api-core >=1.31.5,<2.0.0 || >2.3.0,<3.0.0dev
│   ├── protobuf >=3.15.0,<5.0.0dev (circular dependency aborted here)
...
```
- Remove `zeroconf` dependency. We're using it trough `homeassistant` and want to use the same version as they pin.
- Bump minimum HA version to 2022.6.0
- Relax requirement on `homeassistant-stubs` (add `^`). This allows to update both dependencies by running `poetry add homeassistant@2022.6.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ data:
 
 ### Prerequisites
 
-Use Home Assistant v2022.2.0 or above.
+Use Home Assistant v2022.6.0 or above.
 
 ### Google Account security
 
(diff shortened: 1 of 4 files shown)


leikoilja (approved):
That's an awesome cleanup, thanks :)

PR #556, opened 2022-06-27
Automatically update minimum HA version in HACS config

KapJI:
Keep minimum HA version in `hacs.json` in sync with current installed version in poetry. Poetry dependencies should be a single source of truth.

Creating actual commit is a bit more difficult then updating a single file but it's some fun to play with github api. I didn't give it a proper test because I'm quite lazy for this, let's test in prod :) If it doesn't work, I'll submit a fix.


leikoilja (approved):
Looks cool! Some serious levels of automation we are getting in here 😄  Thanks, man!
I wonder if it will work on the first try :P

PR #559, opened 2022-06-28
Remove empty docs

KapJI:
I don't see a value in having empty docs. They can be restored once there is something useful added.

@leikoilja you may need to disable Pages in repo settings.


leikoilja (approved):
this was a good endeavour, but felt flat 😄

PR #564, opened 2022-06-30
Update README.md

GaworskiRafal:
+ Polish Translation


KapJI:
Can you add it to #563?

leikoilja:
Just as @KapJI mentioned, please add this README change to #563 instead

Issue #562, opened 2022-06-29
HA BETA - 2022.7.0b0 - Deprecation Warnings


jazzyisj:
## Describe the bug

Depracation warnings in HA v2022.7.0b0

## Version of the integration

1.9.13

## Logs

`2022-06-29 19:45:05 WARNING (MainThread) [homeassistant.components.number] custom_components.google_home.number::AlarmVolumeNumber is overriding deprecated methods on an instance of NumberEntity, this is not valid and will be unsupported from Home Assistant 2022.10. Please report it to the custom component author.`

leikoilja:
Thanks for the report, @jazzyisj 💥

KapJI:
List of deprecated methods: "async_set_value", "max_value", "min_value", "set_value", "step", "unit_of_measurement", "value". I guess we should use `_attr_*` instead.

@DurgNomis-drol @ArnyminerZ anyone is interested in fixing this?

PR #569, opened 2022-07-09
Stop overriding deprecated methods of NumberEntity

JasperS2307:
Fixes: https://github.com/leikoilja/ha-google-home/issues/562

Very minimal changes to get rid of the deprecation warnings regarding overwriting deprecated methods of the NumberEntity. 

Note, [the default values](https://github.com/home-assistant/core/blob/dev/homeassistant/components/number/const.py#L10-L12) for min/max/step are the same but with one decimal, think I prefer defining them explicitly to not depend on the default values stat the same, but would be happy to change it if you prefer the default values.

Expected/tested behaviour: 
- No more deprecation warning show up in the logs
- No changes in behaviour

### DIFF
--- a/custom_components/google_home/number.py
+++ b/custom_components/google_home/number.py
@@ -57,9 +57,11 @@ async def async_setup_entry(
 class AlarmVolumeNumber(GoogleHomeBaseEntity, NumberEntity):
     """Google Home Alarm Volume Number entity."""
 
-    _attr_unit_of_measurement = PERCENTAGE
+    _attr_native_unit_of_measurement = PERCENTAGE
     _attr_entity_category = EntityCategory.CONFIG
-
+    _attr_native_min_value, _attr_native_max_value = (0, 100)
+    _attr_native_step = 1
+    
     @property
     def label(self) -> str:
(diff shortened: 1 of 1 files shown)


custom_components/google_home/number.py
```diff
@@ -57,9 +57,11 @@ async def async_setup_entry(
 class AlarmVolumeNumber(GoogleHomeBaseEntity, NumberEntity):
     """Google Home Alarm Volume Number entity."""
 
-    _attr_unit_of_measurement = PERCENTAGE
+    _attr_native_unit_of_measurement = PERCENTAGE
     _attr_entity_category = EntityCategory.CONFIG
-
+    _attr_native_min_value, _attr_native_max_value = (0, 100)
```
KapJI, line 62: I prefer it on separate lines.
```suggestion
    _attr_native_min_value, = 0
    _attr_native_max_value = 100
```

KapJI (changes requested):
Please fix CI errors. Check CONTRIBUTING.md for more details.

KapJI (approved):
Thanks!

PR #633, opened 2023-01-22
Fix build badge

KapJI:
Details: https://github.com/badges/shields/issues/8671


leikoilja (approved):
awesome, thanks for picking it and merging dependabot reviews. was deep in my todo's but haven't got there yet 😅

PR #644, opened 2023-02-01
Replace deprecated DEVICE_CLASS with modern enum

### LINKED ISSUE #625: Integration is using deprecated `DEVICE_CLASS_*` constants
## Describe the bug

This custom integration uses deprecated `DEVICE_CLASS_*` constants in its codebase.

The `DEVICE_CLASS_*` constants have been deprecated and replaced in Home Assistant Core 2021.12 (over a year ago). I would highly suggest updating/migrating this integration to the new enums.

For example, for the device classes supported by the `sensor` platform, there is now a `SensorDeviceClass` enum. So if a sensor previously used the `DEVICE_CLASS_ENERGY` constant, it should now use `SensorDeviceClass.ENERGY`. Other platforms (like `binary_sensor`, and `number`) provide similar enumerations for their supported device classes.

The migration thus only consists of replacing constants with an enumeration member and is, therefore, very low impact and should be fairly straightforward.

If I can help resolve any questions regarding this change or migration, feel free to ask or respond to this issue. I'm happy to help!

Kindest regards,

../Frenck

## Version of the integration

```text
all
```

## Logs

```text
n/a
```

KapJI:
Fixes #625

### DIFF
--- a/custom_components/google_home/const.py
+++ b/custom_components/google_home/const.py
@@ -32,9 +32,6 @@ ICON_ALARM_VOLUME_MID: Final = "mdi:volume-medium"
 ICON_ALARM_VOLUME_HIGH: Final = "mdi:volume-high"
 ICON_ALARM_VOLUME_OFF: Final = "mdi:volume-off"
 
-# Device classes
-BINARY_SENSOR_DEVICE_CLASS: Final = "connectivity"
-
 # Platforms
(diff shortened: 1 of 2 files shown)


leikoilja (approved):
Thank you for finding the time to help out with fixes :)

PR #649, opened 2023-02-04
Migrate to Python 3.10

KapJI:
Home Assistant requires Python 3.10 now. Let's migrate as well.
- Update versions of pre-commit hooks.
- Move mypy and codespell configs to 'pyproject.toml'.
- Update dev dependencies.

### DIFF
--- a/.github/workflows/home-assistant.yml
+++ b/.github/workflows/home-assistant.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 env:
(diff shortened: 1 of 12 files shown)


KapJI:
I try to read change log semi-regularly and sometimes I see user reports about legacy stuff we're using. Also sometimes I see something breaks in automated updates of [`homeassistant-stubs`](https://github.com/KapJI/homeassistant-stubs).

Can you update branch protection rule here as well?

PR #656, opened 2023-02-07
Ignore ha

KapJI:
Revert last update from Dependabot and prevent future updates.

Documentation explains how ignore works: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

I think before we used comment on PR to create in-repo setting. This is more explicit and easier managable.


leikoilja (approved):
Lovely!

PR #657, opened 2023-02-10
better explain pt-pt

ViPeR5000:


leikoilja (approved):
Nice, thank you for helping us fix the translation :)

PR #658, opened 2023-02-10
Revert "better explain pt-pt"

KapJI:
Reverts leikoilja/ha-google-home#657

"Username" is a correct term here. And we don't have all these spaces in other translations.


leikoilja (approved):
my oopsie... thanks for quick fix 😮‍💨

PR #670, opened 2023-02-28
Fix potential vulnerabilities and HA checks

KapJI:
- Update `certifi` to fix potential vulnerability reported by Github.
- Update `homeassistant` to include newer version of `cryptography` which fixes another security issue.
- Rearrange keys in manifest, it's verified by Hassfest check.


leikoilja (approved):
awesome, thank you :)

PR #723, opened 2023-05-29
Allow providing master_token during integration setup

Brephlas:
Small changed so that the Integration now asks for master_token in the set-up process and is therefore able to access the Google API again
Tested with my own environment/account and this was working fine.

Integration Set-up process:
ha-google-home

Test timer on my office's google home device:
ha-google-home-timers

Might be a hacky solution but I tried to keep the usual set up process with an optional input field. Couldn't test this as I was getting the bad authentication error as well.


README.md
```diff
@@ -217,6 +217,9 @@ Your second account would not have access to anything other than Google Home, so
 gets compromised, it's not as severe as losing your main google account.
 See more discussion [here](https://github.com/leikoilja/ha-google-home/issues/124#issuecomment-921773489)
 
+### Master token
+Due to authentication issues with google credentials alone it my be required to get the master token separately and provide it during the setup process. This can be done using [this script](https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d).
```
leikoilja, line 221:
```suggestion
### Master token
Due to authentication issues with google credentials alone it my be required to get the master token separately and provide it during the setup process. This can be done using [this script](https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d) or [glocaltokens](https://github.com/leikoilja/glocaltokens#quickstart) package. 
```

custom_components/google_home/translations/de.json
```diff
@@ -6,7 +6,8 @@
         "description": "Hilfe zur Konfiguration findest du unter https://github.com/leikoilja/ha-google-home. Benutze deinen Google Account und ein App-Passwort. Es ist einfacher und sicherer, ein App-Passwort zu generieren und zu nutzen als das eigentliche Passwort. Es hat weiterhin die selben Zugriffsrechte, ist aber einfacher in der Handhabung beim Skripten als das echte Passwort. (https://myaccount.google.com/apppasswords). Falls nicht, dein normales Google Account Passwort sollte funktionieren.",
         "data": {
           "username": "Google Account Nutzername",
-          "password": "Google Account App-Passwort"
+          "password": "Google Account App-Passwort",
+          "master_token": "Google Account Master Token"
```
leikoilja, line 10: is this a valid german translation? oh well, i guess "master token" is going to be the same in all languages 😅

leikoilja (commented):
Awesome, thanks for finding the time for fixing it :)
A small comment on how to make it possibly better. I would love to test it as well before we merge (will try to do it later today/tomorrow), since i have both username/password and master token working on my setup

custom_components/google_home/translations/de.json
Brephlas: Also think it is going to be the same :D 
For german there is no good translation for "token", which is why we normally just use the word "token" in texts/speech

custom_components/google_home/config_flow.py
```diff
@@ -54,6 +54,11 @@ async def async_step_user(
             username = user_input[CONF_USERNAME]
             password = user_input[CONF_PASSWORD]
             session = async_create_clientsession(self.hass)
+            try:
+                master_token = user_input[CONF_MASTER_TOKEN]
+            except KeyError:
+                # no master token provided
+                master_token = None
```
leikoilja, line 61: Bummer, my comment did get submitted while the first review :/ 

Would have been nice to add input validation here.

Something inline with:
- make all 3 values `username`, `password` and `master_token` optional inputs.
- `.get` all 3 values, fallback to `None` 
- validate that either:
  - username and password are provided
  or 
  - master_token is provided
- instantiate client

Brephlas: Sounds good. I've updated the PR regarding this suggestion. 
Still, if an invalid `master_token` is provided, the integration will just finish the set-up process but with no devices configured. This may be a thing which could also be improved. My idea was to check if it possible to retrieve an `access_token`, but not sure if this should be included here since this is part of glocaltokens

README.md
```diff
@@ -217,6 +217,9 @@ Your second account would not have access to anything other than Google Home, so
 gets compromised, it's not as severe as losing your main google account.
 See more discussion [here](https://github.com/leikoilja/ha-google-home/issues/124#issuecomment-921773489)
 
+### Master token
+Due to authentication issues with google credentials alone it my be required to get the master token separately and provide it during the setup process. This can be done using [this script](https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d) or [glocaltokens](https://github.com/leikoilja/glocaltokens#quickstart) package.
```
leikoilja, line 221:
```suggestion
Due to authentication issues with google credentials alone it may be required to get the master token separately and provide it during the setup process. This can be done using [this script](https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d) or [glocaltokens](https://github.com/leikoilja/glocaltokens#quickstart) package. 
```

custom_components/google_home/config_flow.py
```diff
@@ -51,28 +51,58 @@ async def async_step_user(
...
+            except KeyError:
+                # either no username or password provided
+                username = None
+                password = None
+            
+            try:
+                master_token = user_input[CONF_MASTER_TOKEN]
+            except KeyError:
+                # no master token provided
+                master_token = None
```
leikoilja, line 67: this works, but if you want to simplify, i would suggest going for
```suggestion
  username = user_input.get(CONF_USERNAME)
  password = user_input.get(CONF_PASSWORD)
  master_token = user_input.get(CONF_MASTER_TOKEN)
```

custom_components/google_home/config_flow.py
```diff
@@ -51,28 +51,58 @@ async def async_step_user(
...
+
+            if master_token is None and username is None:
+                self._errors["base"] = "missing-inputs"
+                return await self._show_config_form()
```
leikoilja, line 71:
```suggestion
            if not master_token or (not username and not password):
                self._errors["base"] = "missing-inputs"
                return await self._show_config_form()
```

custom_components/google_home/translations/en.json
```diff
@@ -6,13 +6,15 @@
...
+          "password": "Google account app password",
+          "master_token": "Google account master token"
         }
       }
     },
     "error": {
       "auth": "Username/Password is incorrect. If your google account has 2FA enabled please generate app password (https://myaccount.google.com/apppasswords). If authentication still fails refer to https://github.com/leikoilja/ha-google-home#troubleshooting",
-      "pass-len": "Password length is limited to 100 characters by Google Policy (https://support.google.com/a/answer/139399). Please, change your password or use an app password (https://myaccount.google.com/apppasswords)."
+      "pass-len": "Password length is limited to 100 characters by Google Policy (https://support.google.com/a/answer/139399). Please, change your password or use an app password (https://myaccount.google.com/apppasswords).",
+      "missing-inputs": "Please provide either username and password or master token."
```
leikoilja, line 17: since it's a new translation string not present in any other translation files, we might want to google translate it in all other integration supported languages

custom_components/google_home/config_flow.py
leikoilja: yes, that would be great. 
The right place for token validation and credential validation is probably in `glocaltokens`, you are right. 
But as of now, if you check `_test_credentials` method in this package it calls for `/api/GlocaltokensApiClient/async_get_master_token` that calls to get a master token using credentials to validate if credentials worked out. 
We can just expand that functionality and add `GlocaltokensApiClient.async_get_access_tokens` in the same way, which will call underlying `glocaltokens` client to load access tokens. What do you think?

leikoilja (commented):
Awesome stuff! 💪

custom_components/google_home/config_flow.py
Brephlas: Thanks. Didn't know about the get method :D

custom_components/google_home/translations/en.json
Brephlas: Agreed. I'll add them

custom_components/google_home/config_flow.py
Brephlas: Needed to change the check here because this one was not working (always failing if some inputs are left empty)

custom_components/google_home/config_flow.py
Brephlas: Was my initial thought, too. Implemented it this way and already pushed it to the PR :)

Brephlas:
Updated the PR with a validity check for the master token as well as translations (using DeepL + Google Translator)✌️

leikoilja (approved):
Boom, amazing job @Brephlas 🚀 
Thanks for contributing and fixing such a crucial part! 

Let's merge it in and ask peps to help test it out using the `master` branch build. Once it's verified to work - we release a new version 🔥

leikoilja:
@Brephlas, since pre-commit hooks [were failing on your last commit](https://github.com/leikoilja/ha-google-home/actions/runs/5124141921/jobs/9226877553), i have pushed a minor refactoring to fix linting and typing issues. Please see the commit i [just pushed](https://github.com/leikoilja/ha-google-home/pull/723/commits/b98eee820841216bd6bd87608975b5c89a4d61e3) if you agree with the changes :) I would also highly appreciate if you could pull these changes to your setup and verify that it still all works as expected before we merge :) 
Thank you

Brephlas:
Thanks for refactoring. Just did the tests on my setup and everything works as expected :)

leikoilja:
Closes #339

PR #732, opened 2023-06-25
Publishes a docker version with the prerequisites for ha-google-home already installed.

Luttik:
Creates a CD step that automatically augments the docker-image for Home Assistant with the relevant dependencies.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           context: .
           platforms: linux/amd64,
-          file: ./docker/Dockerfile
+          file: ./image-with-prerequisites.dockerfile
           load: true
           tags: >-
             ghcr.io/luttik/home-assistant:test-latest
@@ -54,7 +54,7 @@ jobs:
         with:
           context: .
           platforms: linux/amd64,
-          file: ./docker/Dockerfile
+          file: ./image-with-prerequisites.dockerfile
           push: true
           tags: >-
             ghcr.io/luttik/home-assistant:latest,
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,0 +1,61 @@
+name: CD
+
+on:
+  push:
+    branches:
+      - main
+      - feature/docker-with-prerequisites
+
+jobs:
+  build:
+    name: 🔨 Build the Docker image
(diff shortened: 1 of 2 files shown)


leikoilja:
@Luttik , interesting idea.
Can you explain a bit more what are you doing here?
If i understand correctly, it builds the docker image with the requirements and then uploads them to the docker registry, right? but why do we need it, how is it helpful? 😅

leikoilja:
@Luttik, hmmm, interesting idea.
but i wonder if this is a right approach. this basically builds a new docker image with specific prerequisites for `ha-google-home` integration for people to base their entire home assistant setup on, right?
although it will save some build time for the first time installers, but provides no more further value. `glocaltokens` depends on `grpcio` which is heavy and takes time to install, but already comes with pre-build wheels from homeassistant. `glocaltokens` takes some time to install, but should also be done only once (when adding the integration) or on consecutive updates. 
would adding a whole new docker image not be an overkill?

Luttik:
Hé @leikoilja,

In my mind, the issue is that it doesn't have to be run once. You have to run it every time you update an image, restart the docker (or possibly the machine). 

An additional benefit of this method is that we can update the dependencies in-line with the version of ha-google-home. I thought of it later, but you could easily pre-install ha-google-home in the docker-image completely. Removing all possible errors that a user can make (up-until the configuration of passwords etc.)... And I know from personal experience that there are many 😉.

I feel like we are getting more and more into a scenario where new fans of home automation are trying to maintain one "it just works" home ecosystem as basis (google home is the main one here). Home assistant than comes around to fill the gaps where necessary, seamless integration here seems like the most valuable development of HA that I can imagine.

leikoilja:
@Luttik i still won't agree that it would be up to us to maintain a docker image of HomeAssistant with pre-build deps. I see the value of avoiding as many as possible deps issues by simplifying installation process (for those using docker only), but that also adds extra maintenance and work for us.
 
if the demand for the integration grows big and deps installation causes a big issue, maybe we should consider moving it into the core of HA so it lives close to the source code and comes out-of-the-box? But I am not sure if HA team takes up on more integration into the core and what is the process there :)

maybe the best option is, you create a new docker image repo on github in isolation, setup CI actions there and ship it from there? Then we can add it as a optional help for ARM users and have a link to your new repo in README section of this integration?

PR #772, opened 2023-10-16
Update README.md

Brephlas:
Add instructions for docker container for token retrieval

PR #780, opened 2023-11-03
Add docs for fetching token using container

KapJI:
I messed up with https://github.com/leikoilja/ha-google-home/pull/772

Credits to @Brephlas

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,20 @@ See more discussion [here](https://github.com/leikoilja/ha-google-home/issues/12
 
 Due to authentication issues with google credentials alone it may be required to get the master token separately and provide it during the setup process. This can be done using [this script](https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d) or [glocaltokens](https://github.com/leikoilja/glocaltokens#quickstart) package.
 
+Since there are several issues getting the token reliable on different environments, you can use a docker container which was created solely for this use: <https://hub.docker.com/r/breph/ha-google-home_get-token>. If you choose to use this container, follow the following instructions:
+
+```
+$ docker pull breph/ha-google-home_get-token:latest
+$ docker run -it -d breph/ha-google-home_get-token
+
+# the previous command will return a container id
+# use this id inside the next command which opens a shell inside the container:
+$ docker exec -it <id> bash
+
+# execute the pre-installed script:
+root@<id>:/# python3 get_tokens.py
+```
+
 ### HACS Installation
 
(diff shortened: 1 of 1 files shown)


leikoilja (approved):
Thanks @KapJI for fixing it & @Brephlas for contribution!

leikoilja (approved):

owenvoke:
I had a couple of questions about this Docker image. Is the `Dockerfile` available anywhere? 👀

And also, if this is solely for this purpose, would it make more sense for the `CMD` to run the `get_tokens.py` script, rather than `bash` so you don't have to connect to the image and leave it running in the background in a detached state?

That way it would just be a case of running this, rather than 3 commands.
```shell
docker run --rm -it breph/ha-google-home_get-token:latest
```

Brephlas:
It is now :) https://github.com/Brephlas/dockerfile_breph-ha-google-home_get-token/tree/master

Regarding the second question: I like the idea and played a little bit with this. Unfortunately, it seems it is not possible to run interactive scripts using this method. At least when asking for user input, the script fails. 
So I think we have two options:
1) stick to the current flow
2) change it, but require to pass username and password using `-e` docker environment variables. But I am not sure if I like that one because the credentials are then inside of the history

owenvoke:
Hi @Brephlas, I've just tested it locally (changing `CMD bash` to `CMD python3 get_tokens.py`) and [it seems to work perfectly](https://asciinema.org/a/T3lXiV9usVuDLQrILvOrel009) with `docker run --rm -it [image]`. 🤔 I'm using macOS, and iTerm2.

<details>
<summary>Preview video</summary>

[asciicast](https://asciinema.org/a/T3lXiV9usVuDLQrILvOrel009)

</details>

Also related to this image, I wonder if we could change to using Alpine to reduce the image size. 🤷🏻
```dockerfile
FROM alpine:3.18

WORKDIR /app

RUN apk add py3-pip
COPY ./get_tokens.py .
COPY ./requirements.txt .
RUN python3 -m pip install -r ./requirements.txt

CMD python3 get_tokens.py
```

This reduces the image from `606MB` to `96.7MB` 🙂

I guess `device_id`, `master_token`, and `access_token` would probably all make sense as having configurable by Docker secrets.

Brephlas:
Thanks @owenvoke. Apparently, the issue was sitting in front of the screen, I missed the `-it` parameter 😅 
It now worked in my local tests and I've already updated the image to alpine and pushed it to dockerhub.  
Will update the docs of this repo in a sec

Regarding the docker secrets we can move the discussion to my repo. However, until now I have no experience with docker secrets

PR #784, opened 2023-11-05
🇸🇰 Add Slovak localization

misa1515:


KapJI:
Please also update the list of languages in README.md.

PR #800, opened 2023-11-21
Fix poetry config

KapJI:
New version of poetry started to verify this.
* Authors was missing `>`
* Non standard layout causes `No file/folder found for package google-home`


leikoilja (approved):
at this points feel free to add yourself to authors as we ll if you want to :) thanks for updating it 💪

PR #806, opened 2023-12-07
Allow Refreshing Devices During Deletes or Manually

jmillnik:
- Add an optional `force_refresh` parameter to `delete_timers` and `delete_alarms`
- Add a `refresh_devices` service to allow manual refreshes

Both options invoke `async_request_refresh` on the coordinator and therefore restart the refresh timer to prevent over-requesting devices. They provide a reasonable (and optional) method to immediately update devices when deleted, or to allow users to refresh timers/alarms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,16 +155,16 @@ service: google_home.delete_alarm
 data:
   entity_id: sensor.kitchen_alarms
   timer_id: alarm/47dc1fa0-5ec0-2cc7-9ead-a94b85e22769
-  force_refresh: true
+  skip_refresh: true
 ```
 
 #### Key Descriptions
 
-| Key             | Example                                      | Description                                      |
-| --------------- | -------------------------------------------- | ------------------------------------------------ |
-| `entity_id`     | `sensor.kitchen_alarms`                      | Entity name of a Google Home alarms sensor.      |
-| `alarm_id`      | `alarm/6ed06a56-8a58-c6e3-a7d4-03f92c9d8a51` | ID of an alarm. See alarms description above.    |
-| `force_refresh` | `true`                                       | Boolean to force refreshing Google Home devices. |
+| Key            | Example                                      | Description                                     |
+| -------------- | -------------------------------------------- | ----------------------------------------------- |
+| `entity_id`    | `sensor.kitchen_alarms`                      | Entity name of a Google Home alarms sensor.     |
+| `alarm_id`     | `alarm/6ed06a56-8a58-c6e3-a7d4-03f92c9d8a51` | ID of an alarm. See alarms description above.   |
+| `skip_refresh` | `true`                                       | Boolean to skip refreshing Google Home devices. |
 
 ### Delete timer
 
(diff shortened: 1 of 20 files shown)


KapJI:
> Add an optional `force_refresh` parameter to `delete_timers` and `delete_alarms`

Do we really need this to be configurable? I'm ok with doing it by default, seems reasonable to me.

jmillnik:
> Yeah, makes sense. Let's default to true and name it `skip_refresh` then. Also it will be great if you can add translations to other languages using some online service.

Makes sense will knock this out after work and get it all cleaned up. Just fixed my pre-commit to fix the linting. Running on a windows laptop atm :(

custom_components/google_home/sensor.py
```diff
@@ -87,13 +89,23 @@ async def async_setup_entry(
...
+                SERVICE_ATTR_FORCE_REFRESH
+            ): cv.boolean,  # type: ignore[dict-item]
+        },
         GoogleHomeAlarmsSensor.async_delete_alarm,
     )
 
     platform.async_register_entity_service(
         SERVICE_DELETE_TIMER,
-        {vol.Required(SERVICE_ATTR_TIMER_ID): cv.string},  # type: ignore[dict-item]
+        {  # types: ignore[dict]
```
KapJI, line 103: This one shouldn't be needed.
```suggestion
        {
```

README.md
```diff
@@ -155,14 +155,16 @@ service: google_home.delete_alarm
...
 
-| Key         | Example                                      | Description                                   |
-| ----------- | -------------------------------------------- | --------------------------------------------- |
-| `entity_id` | `sensor.kitchen_alarms`                      | Entity name of a Google Home alarms sensor.   |
-| `alarm_id`  | `alarm/6ed06a56-8a58-c6e3-a7d4-03f92c9d8a51` | ID of an alarm. See alarms description above. |
+| Key             | Example                                      | Description                                      |
+| --------------- | -------------------------------------------- | ------------------------------------------------ |
+| `entity_id`     | `sensor.kitchen_alarms`                      | Entity name of a Google Home alarms sensor.      |
+| `alarm_id`      | `alarm/6ed06a56-8a58-c6e3-a7d4-03f92c9d8a51` | ID of an alarm. See alarms description above.    |
+| `force_refresh` | `true`                                       | Boolean to force refreshing Google Home devices. |
```
KapJI, line 167: Please update docs as well.

custom_components/google_home/services.yaml
```diff
@@ -15,6 +15,12 @@ delete_alarm:
         entity:
           domain: sensor
           integration: google_home
+    skip_refresh:
+      example: false
+      default: true
```
KapJI, line 20: I meant default true for `force_refresh`. For `skip_refresh` it should be false.

KapJI (changes requested):

KapJI (approved):
Thanks!

PR #590, opened 2022-08-16
Bluetooth sensor integrated, showing the nearest Bluetooth device

ChrisAllen1982:

Issue #123, opened 2021-03-29
Support Bluetooth Scan / Scan Results


stboch:
Back at the start of the GHLocalApi several of us where starting to build a triangulation based presence system based on bluetooth results. Would love to help get this working again as we now have access again. 

[Get Scan Results](https://rithvikvibhu.github.io/GHLocalApi/#operation/GetScanResults)
[Initiate Scan](https://rithvikvibhu.github.io/GHLocalApi/#operation/Scanfordevices)

This would likely require creating new entities as each rescan interval potentially identifies new devices, it all depends.

KapJI:
I remember old official `google_home` integration had this feature. Can we just restore that deleted code?

stboch:
Possibly here is the historical branch for reference 
https://github.com/home-assistant/core/tree/a2589f56e1ac87023123dfb98dffc968073c6ca5/homeassistant/components/googlehome

stboch:
It also looks like someone updated https://github.com/ludeeus/googledevices to support tokens in the client calls.

stboch:
Not sure if @Drakulix has seen this component or if you have seen https://github.com/Drakulix/googlehome might be some opporutinity to increase the development effort and combine the projects :)

leikoilja:
Love the idea of introducing a triangulation presence system by use of Bluetooth based devices as you propose, we could definitely do it if someone is interested in taking a lead on that :) 

Oh nice, haven't seen @Drakulix fork. It doesn't seem to be a very active project yet, but it would be definitely worth giving a shot to try merging the project and inviting @Drakulix to combine efforts with us 🚀

Leatherface75:
Any news about this?

leikoilja:
Not really, @Leatherface75. We don't have anyone interested/available in picking it up yet. @Drakulix agreed to help and mentor anyone who would like to take a lead on this :) So if you have skills and interest, we would be happy you join us ;)

Leatherface75:
Sorry i don't have the skills for bigger things just for small changes in already existing code.

Leatherface75:
Something new about this problem yet?

ArnyminerZ:
Well, maybe @Drakulix can help sticking the code from [Drakulix/googlehome](https://github.com/Drakulix/googlehome) to this repo, so we have the same features.

[...]

PR #891, opened 2024-07-18
Add bluetooth tracking

Lash-L:
Making this as a draft as it is still a WIP, but I'm not sure of my immediate time to continue (might be next week before I finish this)

So I wanted to make this so you knew I was working on it and could give any early feedback if you wished.

Relates to #123

Based off of #590 but with some fundamental changes.
1) Removed device class stuff as it felt like a lot of code to add for very little trade off.
2) Instead of showing the closest device ( I could potentially add this back) - it shows the rssi of the devices you specify.

I developed this within HA Devcontainer, so some of HA styling was auto applied, I will remove it before this PR is completely done.

Will also have a PR for adding the ability to specify known hosts incase zeroconf doesn't work, as I had to work my way around that functionality for getting the devcontainer to work.

Still to do:

- [ ] Add ability to track regular bluetooth mac address
- [ ] Add ability to remove bluetooth mac address or irk mac address
- [ ] Seperate Bluetooth polling, 120 is likely far too high for this and something like 5-10 would make more sense so a bluetooth only update interval would be sane imo


Lash-L:
@leikoilja This should be ready for a review now :) Thanks!

Lash-L:
Set it to create a Hascanner so that the information can be used by other integrations! I had to bump some versioning for it to work correctly

Lash-L:
Remove habluetoothscanner as it was going to cause issues posting an advertisement without data

If in the future someone wants to look at that problem again, they can see my branch here: https://github.com/Lash-L/ha-google-home/tree/bluetooth_scanner_obj

leikoilja:
@KapJI , @DurgNomis-drol , @Drakulix, @ArnyminerZ, any of you guys can help review and test this? 
i currently have no google home device nearby to play around with.

9turnbull (approved):

PR #933, opened 2025-01-15
docs(readme): add more helpful readme references

leikoilja:


KapJI (approved):

leikoilja:
@KapJI , any ideas why `uv` starts to fail randomly? 🤔

KapJI:
Looks like in new uv version, format of lock fine has changed slightly. Just commit updated file.

PR #939, opened 2025-03-02
Added notice about passkeys to the readme.

merlinschumacher:
The authentication via username/password or app password fails reliably when the account uses a passkey as 2FA. When a passkey is required, the master token must be used, as the glocaltoken library doesn't support them. See https://github.com/leikoilja/glocaltokens/issues/521


leikoilja (approved):
awesome, thanks for adding this :)

PR #948, opened 2025-06-08
Update pl.json

nepozs:
Polish translation corrections


leikoilja (approved):
Thanks for contributing

PR #964, opened 2025-08-13
Bluetooth tracking

marcusds:
This builds on  #891

* Fixes some issues with newer versions of HA
* Fixes issues with BT devices never being marked away
  * This is kind of hacky, but we have to a clear and then start a new scan to do so
  * Adds a last seen attr to help with this
* **IRK scanning does not appear to be working**
  * This is a big issue. The code seems to be working properly, but I haven't seen a single match from 2 different phones.
  * It seems like the GH's just aren't returning mac's that are random?
  * I am hoping someone else can look into this and see if there is anything I'm missing.
* Validate IRKs on add
* Normalize MACs to allow for multiple formats to be entered


leikoilja:
@KapJI , do you have a capacity to help with the review here?

marcusds:
@Lash-L Were you able to successfully test IRK based tracking in your PR? Its not working for me and would appreciate any guidance.

